PR #2, opened 2017-08-08
Basic Electron development environment

stel:
This PR contains (moved from https://github.com/realm/realm-browser-js/pull/75):

1. Electron development environment in TypeScript with building & debugging in VS Code
2. Basic Electron app including:
 - Main menu
 - Window management
 - Working with system dialogs
3. Packaging as macOS bundle with signing and stuff

I mark it as "In Progress" because I really like using ElectronForge which @mbalex99 did originally and want to try to support it here. However for now we can start from here unless anyone has any objections.


kristiandupont (approved):

kraenhansen (changes requested):
Minor comments and suggestions - biggest issue I find is that `tslint.json` was deleted and a lot of missing `;` was introduced on that behalf.

kraenhansen (changes requested):
Final suggestion - otherwise I'm all happy ;)

PR #4, opened 2017-08-09
Using Realm in renderer process with existing UI

stel:
This PR adds UI from https://github.com/realm/realm-browser-js/pull/75 and uses Realm in render process.

I has UI for opening local realm files, however supports working with synced realms as well.


kraenhansen (approved):
I have a couple of suggestions that I would like to see implemented, but if we want to move fast I see no reasons why this cannot be merged in its current state.

kristiandupont (approved):

Issue #5, opened 2017-08-10
Setup CI


stel:
Run builds & make releases from CI.

PR #25, opened 2017-08-14
Publish releases from CI

stel:
This PR adds CI job that builds and publishes the results to Github releases.

A proposed workflow looks like this:

1. Everytime the version in `package.json` changes in `master` CI drafts a release on Github and publishes results to that draft.
2. All other commits to `master` update the draft and publish new artefacts.
3. Once release is ready it's published manually on Github (by pressing "Publish" button on draft).
4. After draft is released other commits to `master` don't update its artefacts.

All artefacts are always published on Jenkins's job for every commit on any branch.

This workflow allows having the latest artefacts available for download and testing + automates release process. One bad thing - build takes around 15 minutes (only for mac) but it doesn't look like we can do anything with that :(

Fixes https://github.com/realm/realm-studio/issues/5.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,7 +21,7 @@ node('osx_vegas') {
   stage('Publish') {
     withCredentials([[$class: 'StringBinding', credentialsId: 'github-release-token', variable: 'GH_TOKEN']]) {
       sh '''
-        npm run release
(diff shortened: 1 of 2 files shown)


kraenhansen (commented):
Looking good - just a question.

Issue #74, opened 2017-09-11
As a developer I would like to edit a cell


kraenhansen:
This should be a very minimal implementation, using input type="text" fields and parsing & validation:
- [x] string
- [x] float (using parstFloat)
- [x] decimal (using parstFloat)
- [x] integer (using parstInt)
- [x] boolean ("true", "false")
- [x] date (parse using moment)

Limitations that probably cannot be selected through an input field:
- data editing disabled
- linked objects and lists of objects

PR #118, opened 2017-09-15
Allow edit all kinds of cell but List or Objects

IvanCoronado:
This PR adds new feature required in https://github.com/realm/realm-studio/issues/74


bigfish24 (dismissed):
Looks good. Would be nice to support pressing enter/return when editing a value.

bigfish24 (approved):

PR #175, opened 2017-09-28
Edit object references

### LINKED ISSUE #37: As a developer I would like to edit a cell of type "single object reference"
### TODO

- [ ] Change the visual appearance to open up some sort of ["object selection window"](https://github.com/realm/realm-studio/issues/168) (that can be reused whenever a user is prompted to select an object of a specific type), where an object can be selected from all objects of the class expected by the field.
- [ ] Create an issue on extending this feature with an ability to filter the list of objects.
- [ ] Allow the user to "reset" the value to a `nil` state, if the field is optional.
- [ ] Consider any additional validation that might be needed, as suggested by #3
- [ ] Propagate a change of value to the Realm object.

### LINKED ISSUE #168: As a user I want select an object of a specific type
We need a reusable way for users to select an object of a specific type, to implement #37 and #38.

- [ ] This should open as a modal dialog inside the window.
- [ ] When clicking a row in the table, the dialog closes and the object is selected.
- [ ] Should have a text field that supports filtering with a Realm-JS query. 

Note: Perhaps we could simply reuse the browser components.

itsjgf:
Closes #168 and closes #37.

Summary

1. You can edit an object reference by double clicking on it or right click + “Update reference”.
2. When the modal is open, a list with the current possible references appears and the “Ok” button displays “Set to null”.
3. Once you select a new object, the button will change from ‘Set to null’ to ‘Set’. 

Additional changes which are not related to this PR:

1. Table: Enable row selection on click and editing on double click.
2. Row highlighting: Add missing columnIndex.
3. Edit cell: Add missing autofocus.

### DIFF
--- a/src/ui/realm-browser/Cell.tsx
+++ b/src/ui/realm-browser/Cell.tsx
@@ -9,7 +9,7 @@ import { StringCellContainer } from './types/StringCellContainer';
 
 export const Cell = ({
   onUpdateValue,
-  onListCellClick,
+  onCellClick,
   property,
   style,
   value,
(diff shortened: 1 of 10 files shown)


kraenhansen (changes requested):
Minor changes required - and I did a small commit on the branch as well.

kraenhansen (approved):
For the sake of making progress over perfection - let's get this merged.

PR #176, opened 2017-09-28
Fix to make sure to await the download of the __admin Realm.

bigfish24:
Remove the use of ./data path to store __admin Realm. No need for this?


kraenhansen (changes requested):
I don't think we need to make the function `async` as we are not using the `await` keyword.

PR #293, opened 2017-10-20
The Logs Screen now submits a request with a Refresh Token

mbalex99:
Regarding: https://github.com/realm/ros/issues/546

```typescript
this.socket = new WebSocket(url, [`refresh_token=${userRefreshToken}}`]);
```
Unfortunately, the WebSocket API does not allow you to manipulate headers directly. The only thing you _can_ do is either to put the refresh token as a query string or as a custom protocol value.


kraenhansen (changes requested):
I think we should use basic auth instead - as this is more aligned with the intention of the specification: https://stackoverflow.com/a/4361358/503899

kraenhansen (approved):

Issue #236, opened 2017-10-11
Timeout when opening a large realm file


nirinchev:
There seems to be ~30 sec timeout when opening a Realm file, however, for a file with a huge history, this is too short. The progress indicator continues to fill up even after the error message is displayed but when the file is ultimately opened, I get a message it doesn't have a schema. Then, if I try to open the same file again, I get the same very long download.

To test, connect to `13.74.250.130:9080` and try to open `12f55../consecutiveints`.

PR #402, opened 2017-11-13
feat (realms): Remove timeout while opening realm

IvanCoronado:
Remove timeout opening realm objects. Fixes https://github.com/realm/realm-studio/issues/236

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3,417 +3,465 @@
   "version": "1.7.2",
   "lockfileVersion": 1,
   "requires": true,
-  "dependencies":   {
-    "@types/bcrypt":     {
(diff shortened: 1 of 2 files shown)


kraenhansen (approved):
I noticed you comitted in the `package-lock.json` - thats good. As far as I can tell, it's only changing format - not updating package shas, right? The format gets funked up by the release process (https://github.com/realm/realm-studio/blob/master/Jenkinsfile.prepare#L20 due to an issue with a Jenkins plugin).

I think you need to write "Fixes #236" or "Fixing #236" for GitHub / ZenHub to realize that this PR should close the issue - but I'm not sure.

Issue #158, opened 2017-09-26
Export Schemas


astigsen:
Just like in the Mac Realm Browser, you should be able to export the schema definitions for a given Realm in all the different languages supported by the bindings.

PR #428, opened 2017-11-16
Export schemas

IvanCoronado:
Added "Save model definitions" in the bar menu.
"Save model definitions" options will be enabled just when you have focused a realm browser window.
Languages supported JS and Swift.

closes (partially) #158


kraenhansen (changes requested):
There's a lot of TS code in the exporter service that I would like to follow the codestyle of the project.

Also yes - I would like the tests to be passing before merging this in.

kraenhansen (changes requested):
I think it's a little weird that the services/scheme-export has a "src" folder inside it, would it be difficult to just move the code one level up?

kraenhansen (dismissed):
As soon as the tests pass, I believe this should get merged in and released.

kraenhansen (approved):
I made a few changes - anticipating that we soon will be adding more menu items on the different window types.

PR #439, opened 2017-11-21
Sidebar realm table

### LINKED ISSUE #436: Users clicking a realm in the table of realms sees a sidebar
This sidebar should just have an action to open the realm for now.

IvanCoronado:
Single click above rows in realm table (server-administrations) It will open a sidebar panel with the information.
Double click will open a realm browser window.

closes #436


kraenhansen (changes requested):
👍 for generalizing into a reusable component.

kraenhansen (changes requested):
I agree - it has similar styling.

Almost there ...

PR #443, opened 2017-11-22
Improve dmg style

### LINKED ISSUE #295: Users installing via .dmg on Mac wants a more-Realm experience
Add a link to docs and beautify this window. Should I provide a PNG at a particular size for the background? I can provide a mockup if it helps later

screen shot 2017-10-20 at 10 16 37 am

IvanCoronado:
closes #295


kraenhansen (changes requested):
I suggest using a *.png instead of a *.tiff, thats much more common and supported by electron-builder.

kraenhansen (approved):

PR #445, opened 2017-11-22
Delete realm

### LINKED ISSUE #192: Admins wants to delete a realm
Blocked by https://github.com/realm/ros/issues/366.

Users should be warned before they perform this action.

> Before deleting the Realm here, make sure that any / all clients (iOS, Android, Js, etc.) has already deleted the app or database locally. If this is not done, they will try to upload their copy of the database - which might have been replaced in the meantime.

IvanCoronado:
Admins can now delete realms from an object server.

- Double click is needed for opening a realm browser window
doubleclick-realm

- A panel is opened when a realm is selected in the server administration window
click_realm

- Realms can be deleted
delete-realm

- Protected realms (like `/__admin`) can't be deleted
delete-protected-realm

closes #192 and #436


kraenhansen (changes requested):
We need to URI escape that `path` and I suggest using a native message box for alerts and confirmation dialogs that does not require heavy UI as that takes significantly less code to produce.

kraenhansen (commented):
I wonder if we could find a way to align the "remove" v.s. "delete" naming? It's though as delete is a JS keyword.

kraenhansen (changes requested):

kraenhansen (changes requested):

kraenhansen (changes requested):
I believe this might silence some errors from the server.

kraenhansen (approved):

PR #498, opened 2017-12-05
Fixes #392 open a local sync Realm

### LINKED ISSUE #392: As a User I want to be able to open a local synced file on my disk from ROS
Currently I get: 
image

if I try to open a synced file as a local file.

1. I want to open a file from a ROS installation locally on my disk. At least during development or even under production where a ROS realm file can be shipped for inspection.
1. The error message should be clearer.
2. There is a popup with the same message, and then the window (above) as well. One would suffice.

nhachicha:
Fixes #392


kraenhansen (changes requested):
A suggestion, this would also expose the functionality in case the caller knew the local realm was previously synced.

kraenhansen (changes requested):

kraenhansen (changes requested):

kraenhansen (approved):

PR #532, opened 2017-12-12
csv import into open realm

### LINKED ISSUE #462: Users wants to import data from a CSV file into an existing Realm file
The current Mac Browser allows for import (of multiple .csv files - one per class), but it creates a new Realm-file. I think it would be expected that the data can (alternatively) be imported into the Realm file that is opened - added as the correct class if the schemas match.

nhachicha:
fixes #462 
Menu has been refactored as follow

screen shot 2017-12-12 at 22 03 29
screen shot 2017-12-12 at 22 03 55


kraenhansen (approved):
Looking good!

PR #540, opened 2017-12-13
Allow create primary key column when creating a new class in schema

IvanCoronado:
Now when you create a new class you have 3 choices for the PK field:

- Without primary key
nopk

- With a default value (UUID - int)
defaultpk

- With a customized one
custompk


kraenhansen (changes requested):
Okay - I have a few requests.

kraenhansen (approved):

PR #562, opened 2017-12-18
Open context menu when clicking in content grid

### LINKED ISSUE #529: Right clicking below a table with just one row should should open context menu

IvanCoronado:
close #529


kraenhansen (changes requested):

kraenhansen (changes requested):
Up until this point we've been able to use classes for styling only. Please help me understand why we need to use a class (allowDispatchContentMenuClass) to enable behaviour (opening the context menu)?

To me it would be perfectly fine if the context menu worked in the entire table (including the header).

kraenhansen (approved):
Awesome!

PR #590, opened 2018-01-02
Register Realm Studio as editor for .realm files

IvanCoronado:
- Realm files can be opened with Realm Studio
image

- Double click when APP is opened
openfileappopened

- Double click when APP is closed
openrealmappclosed

- File drop in Realm Studio dock
openrealmwithdock

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     ],
     "fileAssociations": {
       "ext": "realm",
-      "name": "REALM",
+      "name": "Realm",
       "isPackage": true
     }
   },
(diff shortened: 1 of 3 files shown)


kraenhansen (dismissed):

kraenhansen (changes requested):
In general I see that we call `openLocalRealmAtPath` (which returns a promise) but we don't handle any rejections that might happen - in future versions of node this will crash the process, so we should surround it with a try-catch if the function calling is async or use `.catch` to handle any errors when calling `openLocalRealmAtPath`.

https://github.com/realm/realm-studio/blob/master/src/ui/reusable/errors.ts#L3 might be useful to display the error. (but might need some adjustments to be callable from the main process).

PR #607, opened 2018-01-10
Add crash reporter

IvanCoronado:
We need to create a [backtrace-electron instance](http://backtrace.io/electron/) and update [this value](https://github.com/realm/realm-studio/pull/607/files#diff-fc703b1d50a13a77c47cc307f3859fa0R4) with the proper URL.

Use `process.crash()` for testing that It's working properly.


kraenhansen:
This is no longer relevant when we have Sentry https://github.com/realm/realm-studio/pull/839

PR #712, opened 2018-03-01
Enable scrollbar if error cannot fit on screen

cmelchior:
A slightly different use case than https://github.com/realm/realm-studio/issues/693

In my case, the error was too long vertically. This PR adds scrollbars if the content overflows both horizontally and vertically. We should probably still word-wrap the error message, but my first attempt at that failed, so better leave it out of this PR

Old:
image

New:
image


kraenhansen (approved):
Holy moly thats a long error message 👍

PR #715, opened 2018-03-01
Disallow selecting optional for lists of model classes

cmelchior:
Fallout from https://github.com/realm/realm-java/issues/5788

It was possible to select an illegal combination of flags/types, i.e. A list of realm model classes are never allowed to be optional.

This PR hides the option when that combination is chosen. To avoid flickering and make it easy to go back the checkbox is just `invisible` but not removed (so modal size doesn't change), and the flag is remembered, but just ignored when creating a list:

Old behaviour:
image

New behaviour:
image

### DIFF
--- a/src/ui/realm-browser/AddPropertyModal/View.tsx
+++ b/src/ui/realm-browser/AddPropertyModal/View.tsx
@@ -19,6 +19,7 @@ export const View = ({
   focus,
   isList,
   isOpen,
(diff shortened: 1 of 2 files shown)


kraenhansen (changes requested):
Two small suggestions

kraenhansen (approved):
Great!

PR #745, opened 2018-03-10
Add support for disabling internal features.

### LINKED ISSUE #718: Hide "Change Endpoint" option within Realm Studio
I had a user mistakenly change this option and then a great deal of time was wasted troubleshooting why they were not able to connect to realm cloud 

https://github.com/realm/raas/issues/711

To me, this feels like one of those times where we clearly designed something for our own usage without thinking about our users

cmelchior:
Fixes #718 

A better solution would probably be to indicate this depending on if a Github login is used that is a member of the Realm organization, but this will also require changes to the server so seems to be a bit out of scope for now.

This is the practical approach, which just requires setting an environment variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,11 @@ If you've checked out this repository, install and start it by running
     npm install
     npm start
 
+Some internal features, only usable by Realm employees, are hidden by default and are only shown if the following
+environment variable is set:
+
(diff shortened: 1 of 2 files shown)


kraenhansen (changes requested):

kraenhansen (approved):
Great! 🎉

Issue #734, opened 2018-03-08
Refactor the names of folders and components to follow the naming conventions


kraenhansen:
These are the guidelines on naming that we could use https://github.com/realm/realm-studio/blob/master/GUIDELINES.md#react

PR #774, opened 2018-04-16
Unify naming schema

cmelchior:
First part of #734 

Unifies the naming scheme used internally, following the recommendations in https://github.com/realm/realm-studio/blob/master/GUIDELINES.md

This PR only covers the initial UI components under `ui/`, but not all of them. Noticably nothing under `ui/resuable` is refactored yet. This is mostly to validate the approach.

Each component should have its own commit, hopefully making it easier to review


kraenhansen (dismissed):
Great start! I think we should also consider renaming the `*Container.tsx` files to `index.tsx` as mentioned here: https://github.com/realm/realm-studio/blob/master/docs/NAMING-EXAMPLE.md#styled-representational-functional-component-wrapped-by-a-container-component

kraenhansen (approved):
Looking good! Thanks for doing this.

Issue #777, opened 2018-05-02
Show Realm types and allow changing that


nirinchev:
Studio portion of https://github.com/realm/realm-sync/issues/2086

We'd like to allow admins to change the type of the Realm to correct simple mistakes and make it more straightforward to use the adapter/loader.

This involves two tasks:
- [ ] Show the type of the Realm in the browser screen
- [ ] Add button/dropdown to change this type (and show a big red warning that this will invalidate previous permissions that were applied to the Realm).

PR #778, opened 2018-05-03
Upgrading a Realm to a "reference" type

kraenhansen:
This will be fixing #777 by adding the type of the Realm as a column in the table of Realms, showing the `realmType` on the sidebar and a button to change it to `reference` if its `full` or undefined. It also moves the button to collapse the sidebar down-left.

## When clicking a Realm with `realmType === 'full'` or undefined `realmType`

When clicking the "Fine-Grained Permissions" link, the users browser opens https://docs.realm.io/platform/using-synced-realms/access-control#overview.

## Clicking the upgrade button

skaermbillede 2018-05-03 kl 16 38 23


cmelchior (commented):
Only minor points. Mostly about wording. I like the new sidebar 👍 

I have some small UX reservations against the button button to upgrade. It feels slightly weird that this action button is placed in a different place than the other actions and that its color is different. But maybe you wanted to make it a secondary action?

Would it make sense to keep the text, rename the button to "Upgrade" and move it down with the others. So:
Open | Upgrade | Delete

cmelchior (approved):

PR #825, opened 2018-06-11
Fix margin missing from checkbox

### LINKED ISSUE #817: Tickbox of "Save credentials" has no margin
image

cmelchior:
Fixes #817 

image

The fix is a bit hackish, but the URL field also have a hover effect which is taken into account when placed on the screen. Something the checkbox do not, so to align them when both have no focus the padding for the checkbox needs to be offset a little.

### DIFF
--- a/src/ui/ConnectToServer/ConnectToServer.scss
+++ b/src/ui/ConnectToServer/ConnectToServer.scss
@@ -97,6 +97,7 @@
   &__SaveCredentials {
     flex-grow: 1;
     font-size: .8rem;
+    padding-left: $spacer * 1.3;
   }
 
   &__SaveCredentials,


kraenhansen (changes requested):
You are fast! :) I although this fixes the issue I would suggest that we fix it using the "reactstrap" way. The root cause of this bug was because I upgraded Reactstrap but I forgot to add the `check` attribute on the `FormGroup` elements used in this form.

See https://reactstrap.github.io/components/form/ on how to setup a form with a checkmark:

    <FormGroup check>
      <Label check>
        <Input type="checkbox" />{' '}
          Check me out
      </Label>
    </FormGroup>

So - I believe the more correct solution would be to wrap our `Label` in a `FormGroup` with a check attribute [here](https://github.com/realm/realm-studio/blob/master/src/ui/ConnectToServer/ConnectToServer.tsx#L99-L108) or perhaps turn `<div className="ConnectToServer__SaveCredentials">` into a `FormGroup` with the same `className` and the `check` attribute. Makes sense?

kraenhansen:
Thanks again for solving this - I think we should try using Reactstrap attributes instead of styling whereever its possible. I went ahead and implemented that on https://github.com/realm/realm-studio/pull/833.

PR #857, opened 2018-06-19
Deselecting a user before deleting it

### LINKED ISSUE #848: Delete user results in an empty screen
Studio: 2.2.0

**To Reproduce**
1) Log into a ROS instance
2) Select Users
3) Select a User and press delete
4) The window now goes blank instead of showing the list of users

image

image

kraenhansen:
This fixes #848.


cmelchior:
I could have sworn I tried this, and it seemed to work until you killed Realm studio and started it again with `npm run dev`. I was considering if I some problems in the tool/build chain somehow.

kraenhansen:
Okay - I think we have an (unreported) bug in the hot reloading, something about an undefined `__dirname` that I havn't yet had time to locate and fix, perhaps that was what you were seeing?
Perhaps there was an old unmounted component that accessed the user "in the background" or something?

PR #867, opened 2018-07-11
Showing realm state size in the RealmsTabel and RealmSidebar

### LINKED ISSUE #812: Users wants to see the size of a Realm file
#### Goal

As a user I would like to see the size of my Realms so I can plan and manage these.
See https://forums.realm.io/t/how-can-i-see-the-size-of-realms-in-realm-cloud/1441.

#### Expected Results

As a minimum I would expect the size of a Realm being displayed when administering a server and selecting a Realm in the table of Realms. Even better would be if I could see (and potentially sort Realms based on) the size as a column in the table of Realms.

#### Dependencies

ROS does not expose the size of the Realms yet - I've created https://github.com/realm/realm-object-server-private/issues/1080 to track this.

#### Version of Realm and tooling

Realm Studio version: 2.0.0
Realm Object Server version: 3.6.0

kraenhansen:
This fixes #812 and https://github.com/realm/realm-studio-private/issues/2.

It can be tested with ROS running from https://github.com/realm/realm-object-server-private/pull/1113.

Whenever the Realms tab is loaded Studio asks the server to report the state size of all realms on the server. After that it requests the latest state size stats known by the server (these may not be fully up-to-date). These state sizes are used when showing the list of Realms. If the size of a Realm is has not been reported (yet) the loading dots will show in the table.

Here's a (rather long) gif showing the UI when new Realms are added to the server.

realm-sizes-optimized-500

Generating the sizes of 1000 very small Realms takes about 300 ms when running the server locally.

### DIFF
--- a/src/services/ros/index.ts
+++ b/src/services/ros/index.ts
@@ -16,9 +16,6 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-import { remote as electron } from 'electron';
-import * as path from 'path';
-
 import * as realms from './realms';
 import * as users from './users';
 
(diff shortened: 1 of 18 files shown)


kraenhansen:
I changed the header from "state size" to "data size" to be more understandable from an end users perspective.

kraenhansen:
After a demo to @roberhofer we decided to simplify the solution to avoid putting unnecessary stress on the server:
- The Realm state size is fetched from ROS from any previous captured stats emitted by the sync server.
- A refresh icon was added to the header from which the user can manually request the sync server to report the realm state sizes and refetch the stats emitted by ROS.

realm-size-on-demand-fast-and-small

kraenhansen:
When the server is running in the Realm Cloud stats are pulled to Prometheus on a 30s delay. And the UX pattern of refreshing the data size statistics broke down - instead I agreed with @kspangsege that we will only primarily on the server keeping these metrics up to date and gradually move towards a solution with more up-to-date sizes and lower latency upon changes.

skaermbillede 2018-08-01 kl 17 08 21

roberhofer:
@kraenhansen : to your comment on 30s refresh rate: Would 30s be not sufficient?

kraenhansen:
@roberhofer I don't think we can use a refresh UX pattern with a spinner if it takes 30+ seconds, then we would need to show a progress bar or something. In the IRL discussion with @kspangsege he feels that with a server with many realms a user can put the server under too much load by refreshing the state size on-demand and I've therefore moved it to the popover of Studio is started with the environment variable that enables internal features.

kraenhansen:
@roberhofer this is not configurable at the moment. I think we should make an effort improving the solution in a different way. Instead of making the recomputation configurable, we should spend time having the recompute happen instantly when a realm changes, with some sort of "backoff period" to prevent it from recomputing on every single change. Also - we should investigate how to make the recompute faster (perhaps have core update an internal counter when the realm changes) that can be read at constant time, instead of having to traverse the state inside the Realm file.

roberhofer:
Agree that this would be a better mechanism.

kraenhansen:
@roberhofer I summed up the current limitations and potential solutions on https://github.com/realm/realm-studio/issues/881.

Issue #907, opened 2018-08-20
Unable to open encrypted files


Janglinator:
**Describe the bug**
As of version 2.6.1 (I previously had 2.4.1), I can no longer open encrypted files.

**To Reproduce**
Double click a realm file.  The screen just whites out.

**Expected behavior**
I am prompted with a field to enter the encryptionKey

**Screenshots**
![image](https://user-images.githubusercontent.com/8138981/44345858-0f916980-a45a-11e8-8156-7593740926b2.png)
![image](https://user-images.githubusercontent.com/8138981/44345863-1324f080-a45a-11e8-89c2-48c5db98c48f.png)

**Versions:**
 - OS: Mac 10.13.5
 - Studio Version 2.6.1

**Additional context**

kraenhansen:
I tried but cannot reproduce this issue, can I have you try copying the realm file from your simulator to your desktop? I don't know exactly how this filesystem is mounted, but I suspect that it might be interfering with the memory mapping Realm is doing.

Do you get the same experience when opening the realm file by clicking the "Open Realm file" on the Greeting window?
![skaermbillede 2018-08-22 kl 11 32 16](https://user-images.githubusercontent.com/1243959/44455793-33da7b00-a5ff-11e8-9a78-d06b5784820c.png)

Would it be possible for you to share the file and its encryption key with me?

Janglinator:
Moving it to the desktop does solve the issue.  I'll share the file in a bit, I can't get it open myself!

Janglinator:
The directories aren't mounted or linked in any way (that I know of), and a typical directory would look like:

/Users/jangles/Library/Developer/CoreSimulator/Devices/69FB2249-E718-426A-8265-C10C71E6FEBE/data/Containers/Data/Application/97CCED1E-B2DB-486E-BDFE-CD87AEE94377/Documents/default.realm

They can definitely get pretty lengthy, but I can open non-encrypted realms fine without moving them to the desktop.  This only occurs with encrypted files.

What's even stranger is that even after downgrading back to version 2.4.1, I still have the same issue.  So either a) I haven't opened an *encrypted* realm in while, and didn't notice this (most of my recent projects have been non-encrypted), or b) the latest version got me into a broken state (not sure if this is possible, just speculating)

kraenhansen:
Please retest this with v2.7.0 of Realm Studio that just got released and share the exception ID here.

Also - thanks for your patience 👍

kraenhansen:
[...]

PR #913, opened 2018-08-22
Implemented Sentry error boundaries

kraenhansen:
When the React component tree experience an unhandled exception the default behaviour is to unmount the entire component tree, which currently results in a white screen (as with #907). Merging this PR will help debug issues by implementing an error boundary that shows that an unrecoverable error has occurred and prompts the user to send us their error event ID which can be used to retrieve more information on what actually happened from Sentry:

skaermbillede 2018-08-23 kl 10 47 26

This is a very unexpected condition it is hard to provide the user with better guidance on how to recover from the issue - it can be anything really ...


Janglinator:
Is there a way for me to grab this commit, or should I just wait for the next release to try it out on #907?

kraenhansen:
@Janglinator sorry for the late reply on my end - I'll make a release with this now.

Issue #683, opened 2018-02-06
Show object level permissions in Studio


astigsen:
When browsing Realms, you should be able to see (and edit) the permissions on the individual objects. 

The UI could be similar to how we can currently see Realm-level permissions (a sidebar that shows up when you select a row).

This will help users understand the permission model, and will also help our own testing and debugging as we build out the object-level permission feature.

roberhofer:
@kraenhansen : is there actually a API that is available to you to expose this in Studio?

kraenhansen:
The permissions are determined by the __* classes available in the Realm when opened in the query-based mode, so yes - I think I have the API available to expose this in Studio.

bmunkholm:
@kraenhansen What's current estimate on the amount of work left on this?

kraenhansen:
For individual objects specifically, I expect to have something ready for review end of day Wednesday and I expect to have a more polished feature end of week.

kraenhansen:
For the interested I've created an "in active progress" PR with screenshots of the current state of this feature: https://github.com/realm/realm-studio/pull/946.

roberhofer:
Very Cool. @ianpward : have a look please at https://github.com/realm/realm-studio/pull/946

PR #946, opened 2018-09-13
Fine-grained permissions sidebar

kraenhansen:
This PR is adding a first iteration on the ability to view and edit fine-grained permissions of a reference Realm used for query-based sync.

This will be fixing #683 and https://github.com/realm/realm-studio-private/issues/8.

It adds the following functionality:
- [x] List and click to edit permissions already added to an object, class or Realm.
- [ ] ~Add new and remove existing permissions from an object, class or Realm.~
- [ ] ~Click on a role to see, add to or remove from the list of members (this should be virtualized and filterable as it might contain a lot of users).~
- [ ] ~Show when a permission is "shadowed" by a (missing) permission on a higher level.~

☝️ the unchecked items has been moved to [separate issues](https://github.com/realm/realm-studio/issues/1047).

## Initial view with no objects selected
skaermbillede 2018-09-13 kl 15 06 39

## Clicking the `everyone` role
skaermbillede 2018-09-13 kl 15 06 53

## Selecting an `Item` row
skaermbillede 2018-09-13 kl 15 07 09

## Hovering over a privilege shows an explanation
skaermbillede 2018-09-14 kl 09 35 21

## Clicking a "user role"
skaermbillede 2018-09-13 kl 15 07 19

## Selecting a row in a class without object-level permissions
skaermbillede 2018-09-13 kl 15 07 28

For future improvements:
- List and edit permissions on the Realm-level from the list of Realms accessed via the Server Administration window, instead of within the Realm Browser.
- Translate user IDs into human understandable information (by pairing it with account information from the `__admin` Realm).
- An ability to view permissions from a particular users point of view.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,8 +13,9 @@ data/
 # Secrets
 secrets.sh
 
-# Test results from JUnit
+# Test results from JUnit and screenshots from failures
 test-results.xml
+failure-*.png
 
 # Realm locks and management files
 *.realm.lock
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,6 +51,8 @@ jobWrapper {
               keepLongStdio: true,
               testResults: 'test-results.xml'
             )
+            // Archive any screenshots emitted by failing tests
+            archiveArtifacts 'failure-*.png'
           }
         }
       }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -52,7 +52,10 @@ jobWrapper {
               testResults: 'test-results.xml'
             )
             // Archive any screenshots emitted by failing tests
-            archiveArtifacts 'failure-*.png'
+            archiveArtifacts(
+              artifacts: 'failure-*.png',
+              allowEmptyArchive: true,
+            )
           }
         }
       }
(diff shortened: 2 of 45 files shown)


cmelchior:
Note that it makes sense to show all privileges at all levels since values set at lower levels are filtered by values at the higher level, e.g. `canDelete` doesn't have a meaning at the Realm level, but if not set to `true` there, any value you set at the class level will just be ignored.

cmelchior:
Also showing members of a Role the way it is done in the screenshot is dangerous since you can have 1000's of users. We probably need a more advanced popup of some kind, or perhaps just redirect to the __User table with the correct filter enabled (although that would probably also require some nice way to navigate back to the original view.

bmunkholm:
You likely considered making the permission view in the buttom of the left bar (as a collapsible view) ? 
For the number of permissions on this screen it would be fine. But perhaps not if you have many classes and roles.

ianpward:
This looks great! Agree with this:
> Translate user IDs into human understandable information (by pairing it with account information from the __admin Realm).

Also I think it would be good to translate our internal classes from __Permission to Permission or similar

PR #951, opened 2018-09-20
Prepare for a new realm-js upgrade

nirinchev:
This will show a compatibility prompt for users connecting to ROS < 3.11.0 and redirect them to the downgrade page.


src/ui/ServerAdministration/index.tsx
```diff
@@ -86,6 +87,14 @@ class ServerAdministrationContainer
   /* A list of object schemas to use when creating the next Realm */
   protected createRealmSchema?: Realm.ObjectSchema[];
 
+  private readonly compatibilityVersions: Array<{
+    rosVersion: string;
+    studioVersion: string;
+  }> = [
+    { rosVersion: '2.0.0', studioVersion: '1.19.2-ros2' },
```
kneth: Can we have a more recent Studio or do we really have to go back in time?

src/ui/ServerAdministration/index.tsx
```diff
@@ -335,23 +348,35 @@ class ServerAdministrationContainer
...
+      if (compareVersions(version, compatibilityVersion.rosVersion) === -1) {
+        compatibleVersion = compatibilityVersion.studioVersion;
+        break;
+      }
+    }
+
+    const baseUrl = `https://studio-releases.realm.io/v${compatibleVersion}`;
+    switch (os.platform()) {
+      case 'darwin':
+        return `${baseUrl}/download/mac-dmg`;
```
kneth: `let downloadBaseUrl = baseUrl + '/download/'`?

src/ui/ServerAdministration/index.tsx
nirinchev: Not sure I understand? ROS < 3.0 needs that ancient studio version. Or do you propose that we do a one-off studio release with a ROS 2.x compatible realm-js? (I also noticed this should be 3.0.0, not 2.0.0).

nirinchev: If I understand correctly, that's what we're doing - ROS versions below 3.11.0 will prompt the user to downgrade to Studio 2.8.2 which was supposed to be the pinned equivalent of 2.8.0 before I realized there are unreleased changes in master 😓 Then if the ROS version is lower than 3.0.0, we'll prompt the user to downgrade to 1.19.2.

kneth: :+1:

kneth (dismissed):

kneth (approved):

package.json
```diff
@@ -133,8 +133,10 @@
   },
   "dependencies": {
     "@sentry/electron": "^0.8.1",
+    "@types/compare-versions": "^3.0.0",
     "@types/papaparse": "^4.1.31",
     "classnames": "^2.2.5",
+    "compare-versions": "^3.4.0",
```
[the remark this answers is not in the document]
nirinchev: I did not 😓 I didn't realize we were bringing it in as a dev dependency.

package.json
```diff
@@ -133,8 +133,10 @@
   },
   "dependencies": {
     "@sentry/electron": "^0.8.1",
+    "@types/compare-versions": "^3.0.0",
```
kraenhansen: @nirinchev I just noticed this - I think the @types/* packages should be devDependencies, any reason that was not the case for this package?

nirinchev: You're right - I must have fat-fingered the install command, sorry.

src/ui/ServerAdministration/index.tsx
```diff
@@ -335,23 +348,35 @@ class ServerAdministrationContainer
...
-    } else if (platform === 'linux') {
-      return `${baseUrl}/download/linux-appimage`;
-    } else if (platform === 'win32') {
-      return `${baseUrl}/download/win-setup`;
-    } else {
-      return baseUrl;
+  protected downgradedStudioUrl(version: string | undefined) {
+    if (!version) {
+      version = '0.0.0';
+    }
```
kraenhansen: @nirinchev Going over this again, I see that this could have been expressed as a default of the `version` argument:

```
protected downgradedStudioUrl(version: string = '0.0.0') {
```

PR #970, opened 2018-10-30
Cache Realms in Studio

nirinchev:
This changes the renderer behavior to use persistent directory names, allowing it to avoid redownloading the Realms every time a browser window is opened. Additionally, windows are now reused and opening a Realm twice will not open a new window.

studio 2

Finally, a new option to clear the cache is added to the `Help` menu.

screenshot 2018-10-30 at 11 21 11

CM:
For reviewers. On Mac cached files are being stored in: `~/Library/Application Support/Realm Studio/`

### DIFF
--- a/src/main/Application.ts
+++ b/src/main/Application.ts
@@ -35,6 +35,7 @@ import {
   IServerAdministrationWindowProps,
 } from '../windows/WindowProps';
 
+import { cleanupRendererDirectories } from '../utils';
 import { CertificateManager } from './CertificateManager';
 import { CloudManager, ICloudStatus } from './CloudManager';
 import { MainActions } from './MainActions';
@@ -49,9 +50,6 @@ export class Application {
   private windowManager = new WindowManager();
   private certificateManager = new CertificateManager();
   private cloudManager = new CloudManager();
-
-  // Saving a reference for a single greeting window
-  private greetingWindow?: electron.BrowserWindow;
 
   private actionHandlers = {
     [MainActions.AuthenticateWithEmail]: (email: string, password: string) => {
@@ -101,6 +99,11 @@ export class Application {
     ) => {
       return this.showServerAdministration(props);
     },
+    [MainActions.CleanRendererCache]: async () => {
+      await this.windowManager.closeAllWindows();
+      await cleanupRendererDirectories();
+      await this.showGreeting();
+    },
   };
 
   // Instantiate a receiver that will receive actions from the main process itself.
@@ -121,8 +124,6 @@ export class Application {
       // Quit the app if started multiple times
       electron.app.quit();
     } else {
-      // Clean up from any renderer processes
-      this.windowManager.cleanupRendererProcessDirectories();
       // Register as a listener for specific URLs
       this.registerProtocols();
       // In Mac we detect the files opened with `open-file` event otherwise we need get it from `process.argv`
@@ -173,11 +174,17 @@ export class Application {
   }
 
   public async showConnectToServer(props: IConnectToServerWindowProps) {
+    const { window, existing } = this.windowManager.createWindow({
+      type: 'connect-to-server',
+      props,
+    });
+
+    if (existing) {
+      window.focus();
+      return Promise.resolve();
+    }
+
     return new Promise(resolve => {
-      const window = this.windowManager.createWindow({
-        type: 'connect-to-server',
-        props,
-      });
       window.show();
       window.webContents.once('did-finish-load', () => {
         resolve();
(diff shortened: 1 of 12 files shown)


src/main/Application.ts
```diff
@@ -101,6 +99,11 @@ export class Application {
     ) => {
       return this.showServerAdministration(props);
     },
+    [MainActions.CleanRendererCache]: async () => {
```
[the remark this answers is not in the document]
cmelchior: Actually, it seems it should be `ClearRendererCache`, which is why it isn't being called.

src/utils/renderer-process-directory.ts
```diff
@@ -19,39 +19,36 @@
...
-
-export const getRendererProcessDirectories = () => {
-  const directories = fs.readdirSync(userDataPath);
-  return directories
-    .filter(name => rendererPattern.test(name))
-    .map(name => resolve(userDataPath, name));
-};
-
-export const changeRendererProcessDirectory = () => {
+export const changeRendererProcessDirectory = (type: string) => {
```
cmelchior: I'm not sure I understand the rationale for creating a cache directory for each window?

src/ui/reusable/RealmLoadingComponent/index.tsx
```diff
@@ -198,30 +202,33 @@ export abstract class RealmLoadingComponent<
         // Other errors, propagate it.
         throw error;
       }
-    } else if (realm && realm.mode === realms.RealmLoadingMode.Synced) {
-      const props = (realm as any) as realms.ISyncedRealmToLoad;
+    }
+
+    if (realm && realm.mode === realms.RealmLoadingMode.Synced) {
```
cmelchior: I'm not sure this works as expected, but It is unclear to me why not.
I tried to test:
1) Clearing all caches
2) Open the default realm from https://cmelchior.us1.cloud.realm.io. This took ~5 seconds.
3) Close the browser window
4) Open the default Realm again. I would expect this to be instantaneous since the file should be cached, but it it still takes ~5 seconds.

cmelchior (commented):
I fixed a bug which meant that the "Clear Cache" menu item didn't work.

Re-focusing the original window seems to work 👍 
Accessing a cached file is still very slow, so I suspect something is not working correctly?

src/main/MainMenu.ts
```diff
@@ -24,7 +24,7 @@ import * as raas from '../services/raas';
 import { store } from '../store';
 import { showError } from '../ui/reusable/errors';
 
-const isDevelopment = process.env.NODE_ENV === 'development';
+const isDevelopment = process.env.NODE_ENV === 'development' || true;
```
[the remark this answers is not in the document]
nirinchev: Oops, it does - I added it as launching via `npm run start` didn't set the env variable correctly, and must have forgotten to remove it 🤦‍♂️

src/main/Application.ts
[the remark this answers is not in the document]
nirinchev: Thanks! I decided I don't like the names just before comitting and must have forgotten it.

src/utils/renderer-process-directory.ts
nirinchev: We can't have multiple processes accessing the same file/sync session. This is problematic because each renderer runs in its own process, and the server administration window opens the admin Realm. This means that if you open the admin Realm in a browser window, you'll get a realm-js crash.

cmelchior: Ah okay. I thought that everything ran in the same process/thread :+1:

src/ui/reusable/RealmLoadingComponent/index.tsx
nirinchev: Yeah, you'd need to have a huge Realm to notice the difference as the 5 seconds are rendering, logging users in (since we don't serialize right now), etc. If you have a Realm that is several hundred megabytes, then it's quite noticeable.

src/main/Application.ts
```diff
@@ -291,17 +303,24 @@ export class Application {
   }
 
   public showServerAdministration(props: IServerAdministrationWindowProps) {
+    // TODO: Change this once the realm-js Realm.Sync.User serializes correctly
```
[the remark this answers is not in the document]
nirinchev: https://github.com/realm/realm-studio/issues/972

cmelchior (approved):

kraenhansen:
Refocusing the window instead of opening the same Realm in two different windows could be considered an anti-feature. Now a user cannot view+edit objects from multiple classes at the same time by opening the same Realm twice.

I also think the previous behaviour was a good way to show change propagation. A user must now use a separate app or computer to see changes propagate from Studio via the server and back.

nirinchev:
I believe viewing multiple classes at the same time is better implemented as tabs in the same renderer - it's quite inefficient to download a huge Realm to be able to open different views.

I don't think it has been historically obvious to users that they'll be downloading the Realm every time they open it, so even though you could have used it to verify that changes propagate to the server and back, I'd be surprised if anyone depended on it (as without knowing the internals, I would assume that both windows are showing the same local Realm).

kraenhansen:
@nirinchev that makes sense 👍 I think you're right about the tabs, perhaps we should add that to the idea backlog. And you might be right about the users expectations of server roundtrips, I'm probably just a bit blinded about how it's been working so far :)

src/main/Application.ts
```diff
@@ -315,38 +334,51 @@ export class Application {
...
+    });
+
+    let authPromise: Promise<raas.user.IAccountResponse>;
+    if (existing) {
+      window.focus();
+      // authPromise is set in the else clause
+      authPromise = (window as any).authPromise;
+    } else {
+      // Hacky way to avoid recreating the auth promise every time the window is focused.
+      (window as any).authPromise = authPromise = new Promise<
```
kraenhansen: @nirinchev ... why did we do this again? It's super hacky as the comment states and very potent for a future bug.

kraenhansen: If the `authPromise` variable gets renamed or another part of the code does not respect this implicit interface it's ripe for a runtime error.

nirinchev: Sure, that's why there's a comment on both places it's being used - I wouldn't be worried about people renaming the property in one case and not the other. The reason I marked it as a hack wasn't because I add a runtime property on the window object but because it felt semantically incorrect to load the window object with that information. But at the time I couldn't find a better place to persist that information as the architecture didn't seem to follow the MVVM/MVC patterns I am familiar with.

Issue #972, opened 2018-10-30
Serialize/Deserialize users instead of logging them in


nirinchev:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when Studio logs in my user for every window I open.

**Describe the solution you'd like**
I'd like it to use the new serialize/deserialize methods exposed by realm-js.

PR #973, opened 2018-10-30
Serialize users rather than logging them in again and again

nirinchev:
Fixes https://github.com/realm/realm-studio/issues/972


src/ui/ServerAdministration/index.tsx
```diff
@@ -216,7 +216,7 @@ class ServerAdministrationContainer
         },
       });
       // Authenticate towards the server
-      const user = await ros.users.authenticate(this.props.credentials);
+      const user = Realm.Sync.User.deserialize(this.props.user);
```
kraenhansen: As this is synchronous perhaps we could just skip the "authenticating" status?

kraenhansen (approved):
This is pure awesomeness!

PR #987, opened 2018-11-12
Adding a RELEASENOTES.md to the repository + adjusted the prepare script

kraenhansen:


docs/RELEASING.md
```diff
@@ -24,14 +24,15 @@ what it's basically doing:
 1. Checkout the branch
 2. Run `npm version` with the next version specified
 3. Save that new version from the `package.json` into the `package-lock.json`
-4. Commit in the two files changed
-5. Tag the commit with the new version
-6. Waiting for you to confirm the change - see
+4. Append the contents of RELEASENOTES.md into the CHANGELOG.md and restores the RELEASENOTES.md from a template.
```
nirinchev: There's a markdown trick where you can just set all numbers to 1 and it will generate incremental list, so you don't have to update everything when you want to insert something in the middle. E.g. this:

```
1. Foo
1. Bar
1. End
```

renders to

1. Foo
1. Bar
1. End

nirinchev (approved):
Looks good to me

docs/RELEASING.md
kraenhansen: Yeah - I know, thanks :) But my CDO likes the numbers to increment in the .md source too.

PR #989, opened 2018-11-12
Hide path-level permissions block for Query-based Realms

cmelchior:
Previous behavior: The `User permissions` block was shown for all Realm types. This was confusing as the Path-level permissions are not applicable to Query-based Realms.

New behavior: Hide `User Permissions` for `Reference` and `Partial` type Realms. Also push the buttons to the bottom of the screen, always.

We might want to consider showing the Realm-level permissions for `Partial` Realms in this block, but doing that is outside the scope of this PR.

---
image

image

image

image


kraenhansen (approved):
Looking good.

Perhaps it would make sense to add a note which making it obvious that user permissions are managed from within the Realm? I don't think that should hold this change back.

PR #1000, opened 2018-11-14
TSlint updated and errors fixed

kraenhansen:


src/services/data-importer/csv/CSVSchemaGenerator.ts
```diff
@@ -52,8 +52,10 @@ export class CSVSchemaGenerator extends SchemaGenerator {
         schema.properties[header] = Util.isBoolean(value)
           ? 'bool?'
           : Util.isInt(value)
-            ? 'int?'
-            : Util.isDouble(value) ? 'double?' : 'string?';
+          ? 'int?'
+          : Util.isDouble(value)
```
nirinchev: 😱

kraenhansen: I didn't write this :D I just ran the linter :P

nirinchev: See - even the linter has no idea what to do with the conditional 😂 I'm telling you, we should use plain old if-s 🙄

kraenhansen: Agree - for this a plain old if would probably be the best choice :)

PR #1018, opened 2018-11-23
Expose UI for blacklisting users

nirinchev:
Fixes https://github.com/realm/realm-sync/issues/2579

The updated UI:

screenshot 2018-11-23 at 14 44 08

Toggle open:

screenshot 2018-11-23 at 14 50 16

Error message when trying to login in Studio with a blacklisted user:

screenshot 2018-11-23 at 14 44 36

### DIFF
--- a/src/services/ros/index.ts
+++ b/src/services/ros/index.ts
@@ -36,7 +36,22 @@ export type User = IUser & Realm.Object;
 
 export enum UserStatus {
   active = 'active',
-  blacklisted = 'blacklisted',
+  suspended = 'suspended',
+}
+
+export function humanizeUserStatus(userStatus: UserStatus | string) {
+  switch (userStatus) {
+    case UserStatus.suspended:
+      return 'Suspended';
+    case UserStatus.active:
+    case '':
+    case null:
+    case undefined:
+      // We don't have a migration, so assume that every User that doesn't have a status set is Active
+      return 'Active';
+    default:
+      return userStatus;
+  }
 }
 
 export interface IAccount {
--- a/src/services/ros/index.ts
+++ b/src/services/ros/index.ts
@@ -39,16 +39,6 @@ export enum UserStatus {
   blacklisted = 'blacklisted',
 }
 
-export function humanizeUserStatus(userStatus: UserStatus) {
-  switch (userStatus) {
-    case UserStatus.blacklisted:
-      return 'Blacklisted';
-    case UserStatus.active:
-    default:
-      return 'Active';
-  }
-}
-
 export interface IAccount {
   provider: string;
   providerId: string;
--- a/src/services/ros/index.ts
+++ b/src/services/ros/index.ts
@@ -29,9 +29,25 @@ export interface IUser {
   isAdmin: boolean;
   accounts: IAccount[];
   metadata: IUserMetadataRow[];
+  status: UserStatus;
 }
 
 export type User = IUser & Realm.Object;
+
+export enum UserStatus {
+  active = 'active',
+  blacklisted = 'blacklisted',
+}
+
+export function humanizeUserStatus(userStatus: UserStatus) {
+  switch (userStatus) {
+    case UserStatus.blacklisted:
+      return 'Blacklisted';
+    case UserStatus.active:
+    default:
+      return 'Active';
+  }
+}
 
 export interface IAccount {
   provider: string;
--- a/src/ui/ServerAdministration/UsersTable/UserSidebar/UserStatusRoleTable.tsx
+++ b/src/ui/ServerAdministration/UsersTable/UserSidebar/UserStatusRoleTable.tsx
@@ -88,21 +88,19 @@ export const UserStatusRoleTable = ({
             toggle={toggleStatusDropdown}
           >
             <DropdownToggle caret={true}>
-              {humanizeUserStatus(selection.user.status)}
+              {ros.humanizeUserStatus(selection.user.status)}
             </DropdownToggle>
             <DropdownMenu right={true}>
-              {[ros.UserStatus.active, ros.UserStatus.blacklisted].map(
-                status => {
-                  return (
-                    <DropdownItem
-                      key={status}
-                      onClick={() => onStatusChanged(status)}
-                    >
-                      {humanizeUserStatus(status)}
-                    </DropdownItem>
-                  );
-                },
-              )}
+              {[ros.UserStatus.active, ros.UserStatus.suspended].map(status => {
+                return (
+                  <DropdownItem
+                    key={status}
+                    onClick={() => onStatusChanged(status)}
+                  >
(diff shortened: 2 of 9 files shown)


cmelchior:
It seems a bit weird to use a different terminology for the popup and the UI: `Disabled` vs. `Blacklisted` 

Personally I like `Disabled` more since it doesn't sound as aggressive as Blacklisted and I suspect  you might want to disable users for other reasons than being malliocious which blacklisted suggest is the only use case, but that is just bikeshedding so 🤷‍♂️ 

But at least keep them consistent 😄

bmunkholm:
Some thoughts:
The Blacklisting feature is likely going to be rarely used? Should it then be so prominent or "easy" to blacklist? Perhaps it should be a button next to Delete - "Disable" / "Enable"?

Should the wording be "User Enabled/Disabled" instead of "Active/Blacklisted"? I was wondering what "active" means.

cmelchior:
> The Blacklisting feature is likely going to be rarely used? Should it then be so prominent or "easy" to blacklist? Perhaps it should be a button next to Delete - "Disable" / "Enable"?

You could use the same argument for switching between admin or not which is probably just as rare. Personally, I find it quite intuitive the way it is laid out in the PR.

src/services/ros/index.ts
```diff
@@ -29,10 +29,26 @@ export interface IUser {
...
+  blacklisted = 'blacklisted',
+}
+
+export function humanizeUserStatus(userStatus: UserStatus) {
+  switch (userStatus) {
+    case UserStatus.blacklisted:
+      return 'Blacklisted';
+    case UserStatus.active:
+    default:
+      return 'Active';
```
kraenhansen: I suggest defaulting to the raw `userStatus` as defaulting to "Active" could be really confusing.

src/services/ros/index.ts
```diff
@@ -29,10 +29,26 @@ export interface IUser {
...
 }
 
 export type User = IUser & Realm.Object;
 
+export enum UserStatus {
+  active = 'active',
+  blacklisted = 'blacklisted',
+}
+
+export function humanizeUserStatus(userStatus: UserStatus) {
```
kraenhansen: Would probably move this to the component, as it is concerned with how data is displayed and not used by other components yet.

src/ui/ServerAdministration/UsersTable/UserSidebar/UserStatusRoleTable.tsx
```diff
@@ -0,0 +1,100 @@
...
+  selection: ISelection;
+  roleDropdownOpen: boolean;
+  toggleRoleDropdown: () => void;
+  onRoleChanged: (role: ros.UserRole) => void;
+  onStatusChanged: (status: ros.UserStatus) => void;
+  statusDropdownOpen: boolean;
+  toggleStatusDropdown: () => void;
+}
+
+export const UserStatusRoleTable = ({
```
kraenhansen: I think there is something off with the white spacing in this component, specifically
- between the two dropdown buttons - perhaps these should be `size="sm"`?
- between the "status" label and the dropdown
- because its only two dropdown buttons the tabular layout feels like the button to select the role is a button being horizontally centered a bit off. I am a little troubled by this - I don't have a good idea as to what would fix that 😕 ... and perhaps that should not hold back this PR, but we need to find a good solution for this as I imagine we might want to use this pattern elsewhere.

kraenhansen (changes requested):
Although I have some suggestions, there is no doubt in my mind that you've earned that LinkedIn endorsement for React 🎊

src/services/ros/index.ts
nirinchev: I did it like that because we don't have a migration that will set everyone's status to `Active` and wanted to handle the empty/null case. But I get your point - in the future, if we introduce a new status on the server, people with old studio versions may see confusing info. So I'll explicitly handle the `null` and empty string values as `Active` and default to the raw status for others.

src/ui/ServerAdministration/UsersTable/UserSidebar/UserStatusRoleTable.tsx
nirinchev: Yes, I didn't particularly like the layout of these buttons, but while my React skills are top notch 😎, my css skills are somewhat lacking 😓 If you want to give it 10-15 minutes and try to arrange it nicer, I wouldn't mind, but as you said - it's not such a big deal, and although a little ugly, it does get the job done.

src/services/ros/index.ts
nirinchev: We don't have a migration for statuses so by default everything will be null.

src/ui/ServerAdministration/UsersTable/UserSidebar/UserStatusRoleTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+              )}
+            </DropdownMenu>
+          </ButtonDropdown>
+        </td>
+      </tr>
+    </tbody>
+  </table>
+);
+
+function humanizeUserStatus(userStatus: ros.UserStatus | string) {
```
kraenhansen:
```suggestion
function humanizeUserStatus(userStatus: ros.UserStatus | string = ros.UserStatus.active) {
```

How about making active the default value of the argument instead of listing `'', null and undefined`?

kraenhansen (changes requested):

kraenhansen:
> Ping @kraenhansen - do you want to try and play with the UI or should we leave that for a follow-up PR? Other than that, I think I've addressed your comments.

I'll give it a spin first thing tomorrow.

src/ui/ServerAdministration/UsersTable/UserSidebar/UserStatusRoleTable.tsx
nirinchev: I need to check how defaults behave when you pass in `null`. E.g. would `humanizeUserStatus(null)` set `userStatus` to `active` or will it pass `null`?

kraenhansen:
I comitted a suggestion for a change to the buttons size and making them right aligned to relax the layout a bit:

skaermbillede 2018-11-27 kl 10 14 55

Once we have docs explaining the different user states published, I think we should add a link to that page:

skaermbillede 2018-11-27 kl 10 15 19

src/ui/ServerAdministration/UsersTable/UserSidebar/UserStatusRoleTable.tsx
nirinchev: It appears that only `undefined` will get the default value, but `humanizeUserStatus("")` will set `userStatus` to the empty string and `humanizeUserStatus(null)` will set it to `null`.

kraenhansen (approved):
Looking good 👍

Issue #396, opened 2017-11-13
Users wants to immediately see data when opening a large Realm


kraenhansen:
As per @bmunkholm's comment https://github.com/realm/realm-studio/issues/132#issuecomment-330898576 and on Slack
> We shouldn’t have to wait for the realm to be fully synced before we start showing the (partial) content.

This seems to be blocked by the fact that the `Realm.open` returns a promise that does not resolve before the Realm has been completely downloaded.

The progress of download should be visible for the user.

PR #1028, opened 2018-12-03
Reading Realm sizes from a /__metrics Realm

### LINKED ISSUE #917: As a user I would like to know if Studio is connected to ROS
When Studio opens a synchronized Realm (in the Server Administration or Realm Browser window), it would be great to for a user to see the connection state including a log of temporary client-side errors that might happen.

This got possible after Realm JS released with https://github.com/realm/realm-js/pull/1950 merged.

This will also be fixing https://github.com/realm/realm-studio/issues/650.

### LINKED ISSUE #938: Can't see data sizes of Realms in Studio
**Describe the bug**
After upgrading Studio to 2.8.0, I get the "Data Size" column, but it only shows "?":

screen shot 2018-09-06 at 2 01 31 pm

The realms have all existed for months, so the sizes should be pre-calculated on the server, but maybe there is some relation to just having upgraded Studio (previous version I had did not show sizes)?

**Versions:**
 - OS: Mac
 - Studio Version 2.8.0
 - Realm Object Server version 3.9.11

kraenhansen:
Together with https://github.com/realm/realm-object-server-private/pull/1361 this fixes #938.

It also fixes https://github.com/realm/realm-studio/issues/917, showing the connection state for the and https://github.com/realm/realm-studio/issues/396 for the ServerAdministration window.

It also adds usage of the [react-realm-context](https://github.com/realm/react-realm-context) library to the ServerAdministration window.

I've added a note to the release notes on the dependency on a new version of the server, but since this is fixing a broken feature I don't consider that a breaking change.

## Table of Realms with a single size column

Note how both data (state) and file size is shown in a single column. Hovering either reveals a tooltip so users can determine which is which. Also note the green connection state indicator in the top right corner of the window.

skaermbillede 2018-12-17 kl 14 15 36

## Selecting a Realm with a size

Note both sizes on the same line and the "Recalculate size" button was moved to the controls section of the sidebar.

skaermbillede 2018-12-17 kl 14 15 40

## Selecting a Realm without a size

skaermbillede 2018-12-17 kl 14 15 44


src/ui/ServerAdministration/AdminRealm.ts
```diff
@@ -0,0 +1,49 @@
...
+  withRealm: withAdminRealm,
+} = createRealmContext();
+
+export function getAdminRealmConfig(
+  user: Realm.Sync.User,
+  partialConfig: Realm.PartialConfiguration = {},
+) {
+  const config = user.createConfiguration(partialConfig);
+  // TODO: Simplify this once https://github.com/realm/realm-js/issues/1981 is fixed
+  if (config.sync && config.sync.url) {
```
nirinchev: Is there a chance that this is not a sync configuration? I don't think `user.createConfiguration` can ever return a local-only config.

src/ui/ServerAdministration/MetricsRealm.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * When this metric was emitted.
+   */
+  emitted: Date;
+}
+
+/**
+ * Represents a realm_file_size being emitted from the server
+ * @hidden
```
nirinchev: Why is this `@hidden`? I don't expect Studio publishes API docs.

src/ui/ServerAdministration/AdminRealm.ts
kraenhansen: This if-condition is for type inference, `user.createConfiguration` just returns a `RealmConfiguration` which provides no guarantees on the existence of `config.sync`.

src/ui/ServerAdministration/MetricsRealm.ts
kraenhansen: Copy + paste ;) thanks

src/ui/ServerAdministration/AdminRealm.ts
nirinchev: Seems a bit like bending over backwards to satisfy the type checker. It doesn't make sense for `user.createConfiguration` to give us a local config as local configs don't need a user. So I would rather trust the API and force the type checker to cooperate than to write non-obvious code - e.g. I was legitimately confused when reading this about how this method works to be able to get an empty sync config.

nirinchev: If you can propose a change in the typescript definitions in realm-js to produce objects with correct nullability definitions, I'm fine with that, but ultimately, it's not the job of the code in the studio repo to verify that realm-js's API are behaving correctly.

kraenhansen: I agree - tightening the Realm JS types is the right thing to do.

kraenhansen: I've created https://github.com/realm/realm-js/issues/2163 to track this.

kraenhansen:
I updated the screenshots to match the current state of this PR.

robowen5mac:
Has this change been released? In the version of the app I can download (3.20.1) on a Mac, I cannot see the size columns shown above. I am connecting to a realm hosted ROS and have tried to clean and reinstall the app, but nothing I do seems to show the columns shown above

image

kraenhansen:
> but nothing I do seems to show the columns shown above

These columns will only show if a non-empty `/__metrics` Realm exists on the server.

jeanbaptistebeau:
@kraenhansen does this need manual configuration? Can't find any doc about this. The size column is also not showing up for me, same as @robowen5mac, how can I make it visible?

kraenhansen:
Are you collection to an instance if the Realm Cloud?
If yes, the computation of Realm state size was disabled by default since it was more computationally intensive than we initially expected it to be. It might be possible to enabled for single dedicated server instances, please [create a support ticket if you have a dedicated instance](https://support.realm.io/support/home).

jeanbaptistebeau:
Yes, it's running in Realm Cloud. So there's no way to see the size of Realms?

ianpward:
@jeanbaptistebeau You can use the ProgressNotification API to get the amount of data to be downloaded - 
https://docs.realm.io/sync/using-synced-realms/syncing-data#progress-notifications

jeanbaptistebeau:
@ianpward Oh yeah that will work, thanks!

PR #1062, opened 2019-01-02
Read-Only partial realms

### LINKED ISSUE #1031: People must not be able to edit Partial Realms
Partial Realms are individual to a single device. This means that Sync is allowed to optimize the history in ways that are not compatible with an admin user opening the same partial Realm on the server and making changes to it.

Thus Studio needs to make partial Realms read-only. This should either be implemented directly in Studio or through a special read-only mode in JS.

kraenhansen:
This fixes #1031 by adding a warning when opening partial Realms from the server and opening the browser in a "Read-Only" mode where no classes, properties nor objects can be created, the permissions sidebar and editing data is disabled. If for some reason some UI is not disabled correctly it also monkey-patches the `write` method of the Realm to ensure nothing writes to it.

## Dialog

skaermbillede 2019-01-02 kl 10 26 39

Note how the user can choose to not show this warning again.

## The browser opening a partial Realm in read-only mode

skaermbillede 2019-01-02 kl 11 10 12

Note the hint in the left sidebar, telling the user that it's: Opened as "Read Only"

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -11,3 +11,4 @@
 ## Internals
 - Fixed the Dockerfile used when testing PRs. ([#1057](https://github.com/realm/realm-studio/pull/1057))
 - Removing all existing and future unused locals. ([#1058](https://github.com/realm/realm-studio/pull/1058))
+- Ensuring users open partial Realms in a read-only mode. ([#1062](https://github.com/realm/realm-studio/pull/1062))
--- a/src/ui/ServerAdministration/index.tsx
+++ b/src/ui/ServerAdministration/index.tsx
@@ -50,6 +50,8 @@ import {
 } from './AdminRealm';
 import { ServerAdministration, Tab } from './ServerAdministration';
 
+const DONT_SHOW_PARTIAL_WARNING_KEY = 'dont-show-partial-warning';
+
 export type IServerAdministrationContainerProps = IServerAdministrationWindowProps;
 
 export interface IServerAdministrationContainerState {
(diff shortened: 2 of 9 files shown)


src/ui/ServerAdministration/index.tsx
```diff
@@ -181,17 +183,29 @@ class ServerAdministrationContainer
...
+        );
+        if (realmFile) {
+          if (realmFile.realmType === 'partial') {
+            await this.showPartialRealmWarning();
+            await main.showRealmBrowser({ realm, readOnly: true });
+          } else {
+            await main.showRealmBrowser({ realm });
+          }
+        } else {
+          throw new Error("Couldn't find the Realm in the /__admin Realm");
```
bmunkholm: perhaps mention which realm? Or is that obvious from the context?

src/ui/ServerAdministration/index.tsx
```diff
@@ -321,6 +335,29 @@ class ServerAdministrationContainer
...
 
+  private async showPartialRealmWarning() {
+    const show = localStorage.getItem(DONT_SHOW_PARTIAL_WARNING_KEY) !== 'true';
+    if (show) {
+      return new Promise(resolve => {
+        electron.remote.dialog.showMessageBox(
+          electron.remote.getCurrentWindow(),
+          {
+            message:
+              'You are opening a partial Realm created for a specific device. To ensure other divices do not write to this it will be opened in read-only mode.',
```
bmunkholm: divices -> devices.

But perhaps the wording could be better around "To ensure other divices do not write to this".
Perhaps "client" is a better term than "device"?
Perhaps just:
"You are opening a partial Realm created for a specific client. You can only view the content in read-only mode."

src/ui/ServerAdministration/index.tsx
kraenhansen: You just clicked open on a specific Realm, so yeah - I would say so.

bmunkholm:
@kraenhansen Notice I added some comments.

RELEASENOTES.md
```diff
@@ -11,3 +11,4 @@
 ## Internals
 - Fixed the Dockerfile used when testing PRs. ([#1057](https://github.com/realm/realm-studio/pull/1057))
 - Removing all existing and future unused locals. ([#1058](https://github.com/realm/realm-studio/pull/1058))
+- Ensuring users open partial Realms in a read-only mode. ([#1062](https://github.com/realm/realm-studio/pull/1062))
```
bmunkholm: This has a public effect so shouldn't just be under "internals"?

kraenhansen: You're right - moving this to the bugfix section.

PR #1101, opened 2019-02-20
Report the studio user agent

nirinchev:
That'll help us differentiate between studio and app requests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -832,7 +832,7 @@
     },
     "acorn-jsx": {
       "version": "3.0.1",
-      "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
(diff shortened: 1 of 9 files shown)


kraenhansen (changes requested):
I would like for this to be removed from the `process-directories.ts` into the `renderer.ts` at the same location as we're setting the log level: https://github.com/realm/realm-studio/blob/master/src/renderer.tsx#L73-L80

package-lock.json
```diff
@@ -832,7 +832,7 @@
     },
     "acorn-jsx": {
       "version": "3.0.1",
-      "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
+      "resolved": "http://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
```
kraenhansen: I don't like these changes .. what version of NPM are you using?

kraenhansen (approved):

Issue #1092, opened 2019-02-05
Add ability to export / import from on Realm server to another


pierre-moire:
**Is your feature request related to a problem? Please describe.**
There is no way currently to "close the loop" and export in a format that could be imported later. Schemas must be created manually, permissions with another way and data still another way.

This is painful in multiple cases : 
- when deploying from an environment to another
- when replicating data between servers / realms
- when a backup is needed
- when we need to reset a realm from a previous / alternative version 

**Describe the solution you'd like**
The perfect solution would allow to : 
- export a full realm with : 
  - schema
  - data
  - permissions

...and allow to import the exact same file. 

A export / import option in the menu would be a good solution, and even better if we could select what we need to export (schema/data/permissions) and from where (all classes in a Realm or just selected ones)

kraenhansen:
To export the Realm from a server, we could use the https://realm.io/docs/javascript/latest/api/Realm.html#writeCopyTo API.

Importing a local Realm to a server is less trivial.

PR #1134, opened 2019-03-28
Implemented data exporting to JSON or local Realm

### LINKED ISSUE #15: Export data to different formats
We should support the same formats as in #14.

kraenhansen:
This fixes #15 for JSON and another local Realm and part of #1092, by adding a menu item to save the data of a Realm to a JSON file or a local Realm.

Skærmbillede 2019-03-28 kl  14 04 03

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -3,6 +3,7 @@
 ### Enhancements
 
 - Improved how log entries are displayed in the server adminstration window: The timestamp now displays the date when the entry is not from the current day and the context object gets expanded initially. ([#1131](https://github.com/realm/realm-studio/issues/1131), since v1.20.0)
+- Added a menu item to export data to JSON or a local Realm from the Realm Browser window. ([#1134](https://github.com/realm/realm-studio/pull/1134))
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


RELEASENOTES.md
```diff
@@ -3,6 +3,7 @@
 ### Enhancements
 
 - Improved how log entries are displayed in the server adminstration window: The timestamp now displays the date when the entry is not from the current day and the context object gets expanded initially. ([#1131](https://github.com/realm/realm-studio/issues/1131), since v1.20.0)
+- Added a menu item to export data to JSON or a local Realm from the Realm Browser window. ([#1134](https://github.com/realm/realm-studio/pull/1134))
```
bmunkholm: Perhaps add that when the file is exported as a local Realm it will also be stored optimally without additional free space (aka compressed)

bmunkholm: That should also close a related issue to exporting compressed version of a realm.

bmunkholm: @kraenhansen

kraenhansen: I believe the term is "compacted" and as we are not actually zipping the file or anything.

I committed the following change https://github.com/realm/realm-studio/commit/ac0fec03e0f752ffeb41ea41f48824c839b7c7dd

PR #1147, opened 2019-05-03
Allow more powerful queries for Realms/Users tables

### LINKED ISSUE #1146: Realm filter is useless when a lot of realms start with the same prefix
If I have the following realms:
```
/foo
/foo1
/foo2
...
```

I can't get to the `/foo` Realm using the filter because I can't filter out non-exact matches.

nirinchev:
Fixes https://github.com/realm/realm-studio/issues/1146

Some screens:

Screen Shot 2019-05-03 at 17 55 17
Screen Shot 2019-05-03 at 17 55 34
Screen Shot 2019-05-03 at 17 56 00


kraenhansen (dismissed):

package.json
```diff
@@ -254,10 +253,5 @@
   },
   "engines": {
     "node": "^8"
-  },
-  "husky": {
-    "hooks": {
-      "pre-commit": "lint-staged"
-    }
```
kraenhansen: Not a big fan of this change @nirinchev.
Having a pre-commit hook makes it easier to help contributors remember to run linting.

If you think this was really necessary, I think it should have been a separate PR to make it easier to revert in the future.

nirinchev: I couldn't commit due to an error that I showed you some time ago. While I agree it could have been done in a separate PR, it didn't seem worth the effort. I'll create an issue to track enabling it.

kraenhansen: ☝️ did you create an issue to track enabling it @nirinchev? I can't seem to find it ..

src/ui/ServerAdministration/RealmsTable/RealmsTable.tsx
```diff
@@ -81,8 +83,9 @@ export const RealmsTable = ({
         onElementDoubleClick={onRealmOpened}
         onElementsDeselection={onRealmsDeselection}
         onSearchStringChange={onSearchStringChange}
-        searchPlaceholder="Search Realms"
+        searchPlaceholder="Search Realms (start with ! to write a verbatim realm-js query)"
```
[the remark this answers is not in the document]
kraenhansen: @nirinchev I see now that my comment doesn't make that much sense, as you did in-fact add a tooltip with examples 😄I think this placeholder can be greatly simplified as we have that.

package.json
nirinchev: Of course - https://github.com/realm/realm-studio/issues/1166 - it's the latest issue in the repo 😝

PR #1148, opened 2019-06-04
Add progress labels when opening a Realm

nirinchev:
Adds labels to the Realm download bar:

Screen Shot 2019-06-04 at 13 47 07


cmelchior (approved):

src/ui/reusable/LoadingOverlay/LoadingOverlay.tsx
```diff
@@ -75,7 +76,10 @@ export const LoadingOverlay = ({
             className="LoadingOverlay__Progress"
             value={progress.transferred}
             max={progress.transferable}
-          />
+          >
+            {// tslint:disable-next-line: prettier
```
kraenhansen: @nirinchev: You could move this to a helping function which takes `progress` as an argument, instead of disabling prettier. I think it's the only place in the codebase that prettier gets disabled ... for no good reason in my opinion.

PR #1149, opened 2019-06-04
Assorted dev experience improvements:

nirinchev:
- Persist whether a user is Realm employee to avoid having to run studio with magic env variables
- Add an option to open the cache folder to make it easier to inspect client Realms
- Expose a way to connect to ROS from a custom url.

For the custom url, the format is:
`x-realm-studio://connect-to-ros?url=ROS-URL&credentials=JSON_SERIALIZED_CREDENTIALS`.

For example:

`x-realm-studio://connect-to-ros?url=https%3A%2F%2Ffoo-bar.cloud.realm.io%2F&credentials=%7B%0A%20%20%20%20%22kind%22%3A%20%22jwt%22%2C%0A%20%20%20%20%22providerName%22%3A%20%22jwt%22%2C%0A%20%20%20%20%22token%22%3A%20%22some-long-token%22%0A%7D`

Decoding, we can see the two parameters:
- url: `https://foo-bar.cloud.realm.io/` 
- credentials: 
  ```
  {
      "kind": "jwt",
      "providerName": "jwt",
      "token": "some-long-token"
  }
  ```

### DIFF
--- a/src/main/MainMenu.ts
+++ b/src/main/MainMenu.ts
@@ -24,7 +24,7 @@ import * as raas from '../services/raas';
 import { store } from '../store';
 import { showError } from '../ui/reusable/errors';
 
-const showInternalFeatures =
+const enableTogglingInternalFeatures =
   process.env.NODE_ENV === 'development' ||
   process.env.REALM_STUDIO_INTERNAL_FEATURES === 'true'; // Show features only relevant for Realm employees
 
@@ -48,6 +48,7 @@ function generateCloudEndpointItems(
 export const getDefaultMenuTemplate = (
   updateMenu: () => void,
 ): electron.MenuItemConstructorOptions[] => {
+  const showInternalFeatures = store.shouldShowInternalFeatures();
   const electronOrRemote = electron.remote || electron;
   const template: electron.MenuItemConstructorOptions[] = [
     {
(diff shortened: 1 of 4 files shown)


src/main/MainMenu.ts
```diff
@@ -98,42 +99,52 @@ export const getDefaultMenuTemplate = (
     {
       label: 'View',
       submenu: [
+        {
+          label: 'Show Internal Realm features',
+          visible: showInternalFeatures || enableTogglingInternalFeatures,
```
cmelchior: Any particular reason for wanting to disable the internal features? Testing? Since you special knowledge to enable them in the first place?

src/store.ts
```diff
@@ -31,30 +31,36 @@ class RealmStudioStore {
...
   }
 
   public toggleShowSystemClasses() {
-    const currentValue = this.store.get(this.KEY_SHOW_SYSTEM_CLASSES, false);
+    const currentValue = this.shouldShowSystemClasses();
     this.store.set(this.KEY_SHOW_SYSTEM_CLASSES, !currentValue);
   }
 
+  public toggleShowInternalFeatures() {
+    const currentValue = this.shouldShowInternalFeatures();
```
cmelchior: How does this behave if the value is `undefined`? Is `!undefined === true` ?

cmelchior (approved):

src/main/MainMenu.ts
nirinchev: Right - there's no particularly compelling reason to want to disable internal features - I guess if you're not really using them anymore, you may want to reduce clutter in the menus. And just felt awkward to have a feature you can enable, but not disable 😅

cmelchior: No big issue, I'm fine with keeping it

src/store.ts
nirinchev: Yes, `undefined` is falsy, so `!undefined` is strictly equal to `true`.

src/main/MainMenu.ts
kraenhansen: It's nice to be able to disable the internal features when you want to do a screenshot or screencast showing off a feature to regular users.

Issue #1166, opened 2019-09-23
Reenable lint-staged


nirinchev:
`lint-staged` was disabled in https://github.com/realm/realm-studio/pull/1147 because it was unstable on macos (failing due to missing git). We should reenable it and figure out what was wrong.

kraenhansen:
I decided to not add this back in and removed the dependency instead.
CI will fail if the linter doesn't pass.

PR #1168, opened 2019-09-23
Re-installed and -enabled the husky precommit hook

kraenhansen:
Reverting the disabling of husky as per https://github.com/realm/realm-studio/pull/1147 fixing https://github.com/realm/realm-studio/issues/1166.


kraenhansen:
I decided not to re-enabled this.

PR #1172, opened 2019-09-24
Hash windows paths

### LINKED ISSUE #793: Windows: Support long file paths
See discussion here:
https://realmio.slack.com/archives/CA3ET33NU/p1527116304000354

nirinchev:
Closes https://github.com/realm/realm-studio/issues/793


nirinchev:
@kraenhansen I'm not typically one to debate coding style, but that is the exact opposite of pretty: https://github.com/realm/realm-studio/pull/1172/commits/0ac629308fa834c69b2032fce3ec620a88925c5c#diff-9b5622305354798878e1836ed2d77d9dL90-R104. Do you think we can raise the line length limit to something more meaningful?

src/services/ros/realms.ts
```diff
@@ -82,6 +87,31 @@ export const open = async (params: {
...
+        .update(realmPath)
+        .digest('hex'),
+    );
+
+    fs.ensureDirSync(path.join(process.cwd(), user.identity));
+
+    // tslint:disable-next-line:no-console
+    console.log(
+      `Rewrote: '${user.identity} - ${getUrl(user, realmPath)}' to '${result}'`,
+    );
```
kraenhansen: Do we really need this? I suggest removing it.

src/services/ros/realms.ts
```diff
@@ -82,6 +87,31 @@ export const open = async (params: {
   return realm;
 };
 
+const getPathOnDisk = (
```
kraenhansen: I think it's not trivial why this rewrite is happening - and especially not why its happening only on Windows. Perhaps you could add a comment and optionally a link to the issue on GitHub?

Jenkinsfile
```diff
@@ -101,7 +101,7 @@ pipeline {
       steps {
         nvm(env.NODE_VERSION) {
           // Install dependencies
-          sh 'npm install'
+          sh 'npm ci'
```
kraenhansen: 🥇

kraenhansen (changes requested):
Overall great - is there any good reasons we want to do this only on Windows? It might make it simpler to reason about if we just made it the default on any platform?

kraenhansen:
> @kraenhansen I'm not typically one to debate coding style, but that is the exact opposite of pretty: [0ac6293#diff-9b5622305354798878e1836ed2d77d9dL90-R104](https://github.com/realm/realm-studio/commit/0ac629308fa834c69b2032fce3ec620a88925c5c#diff-9b5622305354798878e1836ed2d77d9dL90-R104). Do you think we can raise the line length limit to something more meaningful?

It's just left at the default (which looks to be 80, judging by https://prettier.io/docs/en/options.html#print-width and https://prettier.io/docs/en/rationale.html#print-width):
> Prettier, on the other hand, strives to fit the most code into every line. With the print width set to 120, prettier may produce overly compact, or otherwise undesirable code.

In principle I don't see an issue with increasing the limit from 80. At the same time I can also see how you could easily break up the lines you reference into multiple lines using intermediary variables with meaningful names. There will always be some lines that hits right about the limit, whatever the limit might be.

What do you think a more meaningful limit would be?

src/services/ros/realms.ts
nirinchev: I think it's useful to know the name of the file that corresponds to a url. An alternative approach I was considering was to add a menu item in the browser to locate the file on disk, but that seemed too much for something so simple. I think it's mostly useful for debugging purposes (e.g. to see how large the client realm is after initial sync), so opening the debug tools and checking the console output seemed like the best tradeoff between preserving the functionality and not overengineering it.

nirinchev:
Re: why only on windows - I still feel the verbose paths are easier to read (if one has to) and considering we don't have limitations on other platforms, I don't see a good reason to rewrite the paths there.

Re: prettier - I think the width of the github code browser is good enough - I think that's around 120 characters. We could also try 100 or 120 and review the changes it makes and see if something becomes overly obnoxious.

src/services/ros/realms.ts
kraenhansen: In its current form Realm Studio is writing pretty much nothing to the console (except for Electron and Realm which writes directly to stdout), hence the no-console lint rule.

In my experience it's easier to instrument the app with console.logs when you're debugging a particular issue (or attach a debugger?) then to graciously put them anywhere they might come in handy. You'll always miss that thing you needed anyway.

Perhaps we should add an easy way to get the size on disk of the client-side Realm instead?

nirinchev: I feel strongly there should be a way for us to reverse map a hashed path to a file on disk. Console.logging the path was the cheapest way to achieve that, considering Realm devs would be able to open the developer tools and inspect the console output. If you prefer to achieve that goal in a different way, feel free to contribute to the PR and expose menu options.

kraenhansen (approved):

jtoselli:
Great enhancement :)

Issue #1165, opened 2019-09-23
Update dependencies


bmunkholm:
- [ ] Electron 7 (we're currently using 4.2.6), but upgrading requires [Realm JS to support Node.js v12](https://github.com/realm/realm-js/issues/2516).

kraenhansen:
Fixed with https://github.com/realm/realm-studio/pull/1274 and https://github.com/realm/realm-studio/pull/1278

PR #1174, opened 2019-09-26
Upgrading to Electron 4

kraenhansen:
This also upgrades `electron-updater`, which we've already had a few issues with. It's important that upgrading from this version of Realm Studio is properly tested on all platforms before getting released.

We might want to hold back on merging this until [Realm JS has support for NAPI](https://github.com/realm/realm-js/issues/2516), consequently NodeJS v12, unlocking [Electron 6 support](https://electronjs.org/blog/electron-6-0).

This needs an upgrade of Realm JS as the Electron 4 binary was not available for v2.23.0 of Realm JS.


kraenhansen:
This (partially) fixes https://github.com/realm/realm-studio/issues/1165

PR #1177, opened 2019-09-26
Upgrading to Realm 3.2.0

kraenhansen:
~This PR is against the `next-major` branch as Realm JS in this version cannot connect to version of ROS below v3.23.1.~ It remains to be tested, but there might be no issues in upgrading Realm JS, in regards to backwards compatibilities against earlier versions of ROS.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,5 +1,9 @@
 [Changes since {PREVIOUS_VERSION}](https://github.com/realm/realm-studio/compare/{PREVIOUS_VERSION}...{CURRENT_VERSION})
 
+### Breaking Changes
+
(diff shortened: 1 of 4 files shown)


nirinchev:
Why does this only work with servers running 3.23.1?

kraenhansen:
> Why does this only work with servers running 3.23.1?

No have no idea - @kneth might be able to answer that? I'm just reading the release notes https://github.com/realm/realm-js/releases/tag/v3.1.0.

kraenhansen:
I had a chat with @kneth IRL and it turns out the compatible version of ROS mentioned in the changelog is not a certain lower bound, but instead the version which the release was tested against. So ... it might be compatible with ROS in the same way that Realm JS 2.23.0 is.

PR #1192, opened 2019-10-17
Fixing documentation urls

kraenhansen:
This fixes the URLs, as per @karl-denby's suggestion on Slack.


karl-denby (commented):
We could remove react-native button+link until someone creates a proper full sync tutorial for that but otherwise this is great

kraenhansen:
@karl-denby do you have a template / guide / format for creating it? What does it need to include? I could probably make one.

karl-denby:
The query based sync example we had previously is [here](https://docs.realm.io/sync/v/3.x/getting-started-1/react-native-quick-start/step-1-query-based-sync).

It would basically need to be the same but with full sync instead.

Issue #309, opened 2017-10-23
Users opening Realm files of an older version, wants a warning and an option to upgrade


kraenhansen:

kraenhansen:
Fixed by https://github.com/realm/realm-studio/pull/1236

PR #1236, opened 2019-12-17
File format upgrade dialog

kraenhansen:
This solves #8 & #309 by adding a dialog prompting the user to choose if they want to upgrade (optionally making a backup first).

upgrade-prompt
upgrade-backup-saved

Will fail if a file exists at the location that the backup would be stored:
upgrade-file-exists

### DIFF
--- a/src/ui/RealmBrowser/index.tsx
+++ b/src/ui/RealmBrowser/index.tsx
@@ -17,6 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 import { ipcRenderer, MenuItemConstructorOptions, remote } from 'electron';
+import fs from 'fs-extra';
 import path from 'path';
 import React from 'react';
 import Realm from 'realm';
@@ -24,7 +25,7 @@ import Realm from 'realm';
 import { DataExporter, DataExportFormat } from '../../services/data-exporter';
 import * as dataImporter from '../../services/data-importer';
 import { Language, SchemaExporter } from '../../services/schema-export';
-import { menu } from '../../utils';
+import { menu, realms } from '../../utils';
 import {
   IMenuGenerator,
   IMenuGeneratorProps,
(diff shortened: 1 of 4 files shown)


src/ui/RealmBrowser/index.tsx
```diff
@@ -295,15 +298,62 @@ class RealmBrowserContainer
...
+      message === FILE_UPGRADE_NEEDED_MESSAGE &&
+      realm.mode === realms.RealmLoadingMode.Local
+    ) {
+      const buttons = ['Cancel', 'Upgrade in-place', 'Backup and upgrade'];
+      const answerIndex = remote.dialog.showMessageBox({
+        type: 'question',
+        buttons,
+        defaultId: 2,
+        title: 'Realm file needs an upgrade',
+        message: 'The Realm file stores data in an outdated format',
```
bmunkholm:
```suggestion
        message: 'The Realm file stores data in an older format not supported by this Studio version.',
```

src/ui/RealmBrowser/index.tsx
```diff
@@ -295,15 +298,62 @@ class RealmBrowserContainer
...
+        detail:
+          'This file needs to be upgraded to a newer file format before it can be opened. Would you like a backup of the file, before performing an irreversible upgrade of the file?',
```
bmunkholm:
```suggestion
          'This file needs to be upgraded to a newer file format before it can be opened by this version of Studio. Would you like a backup of the file, before performing an irreversible upgrade of the file?',
```

bmunkholm: Not sure if we should inform about the option to simply open the file in an older version of Studio. Or is that too obvious?

kraenhansen: We don't necessarily release compatibility versions for every version of the file format - and it's not trivial for a user to determine which compatibility version they should download (or where they are downloaded from). I think we can safely assume that the vast majority of users wouldn't be able to download a version which is compatible with a particular the file format version.

I suggest not adding " by this version of Studio." as the message is already somewhat long and a slightly complicated to understand.

PR #1282, opened 2020-05-11
Removing everything related to ROS and RaaS (legacy Realm Cloud)

kraenhansen:
This removes all and any functionality related to ROS and Realm Cloud, as this won't be relevant in the first iteration of the MongoDB Realm Studio application.


bmunkholm (approved):
Let me put it this way - I can't see anything obviously wrong, but I also haven't tested it :-)
Only missing thing is a release note update.

kneth (approved):
It's a huge PR, and without detailed knowledge I believe it looks good.

PR #1291, opened 2020-05-26
Manually write or generate ObjectId as part og create flow

steffenagger:
## Screenshots of added PK guiding:
image
image


src/ui/RealmBrowser/Content/CreateObjectDialog/index.tsx
```diff
@@ -87,16 +87,18 @@ class CreateObjectDialogContainer extends React.PureComponent<
...
     ) {
-      return uuid();
+      return new ObjectId();
     } else if (property.type === 'list') {
       // If a list is optional, it refers to the type of the elements
       return [];
     } else if (property.optional) {
       return null;
+    } else if (property.type === 'object id') {
+      return new ObjectId();
```
nirinchev, line 101: I agree with this change, but doesn't that make the first condition meaningless? I.e. if we'll generate object id for all properties of `object id` type, then what does it matter if they're called `_id`?

steffenagger: Any optional/nullable property defaults to `null` in the if clause before the one mentioned:
https://github.com/realm/realm-studio/blob/cbcf32d85f5989143b82c3bc1310df47fb466701/src/ui/RealmBrowser/Content/CreateObjectDialog/index.tsx#L98-L99
The first condition opts out of defaulting to a `null` value if the primary key is optional.

I had originally removed the first if clause, until I realised a primary key can be optional/nullable. I'm not sure in what scenario an optional primary key would make sense, and would argue that this option could only end up confusing developers - but it's currently possible.

steffenagger: I guess that actually didn't answer the actual question:
> then what does it matter if they're called _id?

In the first if clause I should probably only check if it's a primaryKey of type `'object id'`.

nirinchev: I see, so if an object id field is nullable, we only want to generate a value if that field is a PK? I guess that's fine although I think it might be a good idea to generate it either way - it's always easier to assign a generated value to null than to assign null value to some random object id.

steffenagger: I've added a “generate button” to the `ObjectIdControl`, for easy random generation of ObjectIds, partly to keep the default nulling of optional values.

src/ui/RealmBrowser/AddClassModal/AddClassModal.tsx
```diff
@@ -136,6 +154,29 @@ export const AddClassModal = ({
...
+                <CardText>
+                  <small>
+                    {
+                      'If the primary key is named "uuid" with the type "string", Studio will auto generate a UUID (v4) for new entries.'
+                    }
+                  </small>
+                </CardText>
+              </CardBody>
+            </Card>
+          )}
```
kraenhansen, line 179: I suggest uploading a screenshot of this to the PR description. It makes it easier for others to review the change in context 👍

src/ui/RealmBrowser/Content/CreateObjectDialog/index.tsx
```diff
@@ -85,18 +86,20 @@ class CreateObjectDialogContainer extends React.PureComponent<
...
-      property.type === 'string'
-    ) {
-      return uuid();
+    if (propertyName === primaryKey) {
+      // Special handling for primary keys. Opting out of optional handling & return a random value.
+      if (property.type === 'object id') {
+        return new ObjectId();
+      } else if (propertyName === 'uuid' && property.type === 'string') {
+        return uuid();
+      }
```
kraenhansen, line 95: I assume we need an else-if statement for all types that are possible as primary keys, in particular I think we need a branch for `int`. To avoid accidentally returning `undefined` I would suggest adding an `else` statement that throws with:

```typescript
throw new Error(`Unexpected type "${property.type}" as primary key`);
```

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
```diff
@@ -0,0 +1,105 @@
...
+export class ObjectIdControl extends React.PureComponent<
+  IBaseControlProps,
+  IObjectIdControlState
+> {
+  constructor(props: IBaseControlProps) {
+    super(props);
+
+    this.state = {
+      internalValue: props.value,
+    };
```
kraenhansen, line 39: Historically we've [initialized the state when defining it as a member](https://github.com/realm/realm-studio/blob/master/src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectControl/index.tsx#L51-L53) - I do however see the benefit that with this approach you don't need to write the type twice.

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
```diff
@@ -0,0 +1,105 @@
...
+        )}
+        {children}
+      </InputGroup>
+    );
+  }
+
+  private inputChangeEventHandler = (e: React.ChangeEvent<HTMLInputElement>) =>
+    this.internalChangeHandler(e.target.value);
+
+  private clearValue = () => this.internalChangeHandler(null);
```
kraenhansen, line 77: Historically these event handlers have been following [a naming scheme prefixing the name of the action with `on`](https://github.com/realm/realm-studio/blob/master/src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectControl/index.tsx#L77):
```suggestion
  private onClearValue = () => this.internalChangeHandler(null);
```

I do see however [the React docs suggests `handle*`](https://reactjs.org/docs/forms.html).

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    if (val) {
+      try {
+        parsedId = parseObjectId(val, property);
+      } catch (_) {
+        // ignored
+      }
+    }
+
+    onChange(parsedId);
```
kraenhansen, line 94: I'm curious if we could limit this to calling `onChange` only when we have a valid parsed object ID or the `val` is missing / empty?

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    this.setState({ internalValue: val });
+
+    let parsedId: ObjectId | null = null;
+
+    if (val) {
+      try {
+        parsedId = parseObjectId(val, property);
+      } catch (_) {
+        // ignored
```
kraenhansen, line 90: I suggest logging this as a `console.warn`, since the error could be something else than a parsing error. Ex an error in the implementation of `parseObjectId`, which would be hard to find when the error is swallowed completely.
I believe we've done that elsewhere.

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
```diff
@@ -0,0 +1,105 @@
...
+      </InputGroup>
+    );
+  }
+
+  private inputChangeEventHandler = (e: React.ChangeEvent<HTMLInputElement>) =>
+    this.internalChangeHandler(e.target.value);
+
+  private clearValue = () => this.internalChangeHandler(null);
+
+  private internalChangeHandler = (val: string | null) => {
```
kraenhansen, line 79: I don't see a need to abbreviate to `val` here.
```suggestion
  private internalChangeHandler = (value: string | null) => {
```

Also not the a fan of the naming of this method, since it's implied that it's internal since its `private`.
I would probably have called it `onChange` like the prop - since its basically just delegating to that method internally (after parsing of course).

RELEASENOTES.md
```diff
@@ -7,7 +7,9 @@
 
 ### Enhancements
 
+- Added support for creating & generating `ObjectId` in create-flow. ([#1291](https://github.com/realm/realm-studio/pull/1291))
```
kraenhansen, line 10:
```suggestion
- Added support for creating & generating `ObjectId` when creating objects. ([#1291](https://github.com/realm/realm-studio/pull/1291))
```

RELEASENOTES.md
```diff
@@ -7,7 +7,9 @@
...
 - Added support for editing existing `ObjectId` values. ([#1290](https://github.com/realm/realm-studio/pull/1290))
+- Primary keys in Class/Schema creation now defaults to prop name `_id` & type `ObjectId`.
```
kraenhansen, line 12:
```suggestion
- Primary keys in Class/Schema creation now defaults to an `ObjectId` property named `_id`.
```

src/ui/RealmBrowser/AddClassModal/AddClassModal.tsx
```diff
@@ -136,6 +154,29 @@ export const AddClassModal = ({
...
           </FormGroup>
+          {primaryKey && (
+            <Card>
+              {/* Todo: Wording */}
+              <CardBody>
+                <CardSubtitle>MongoDB Sync:</CardSubtitle>
+                <CardText>
+                  <small>
+                    {
+                      'If this Realm is intended to be synced with MongoDB, ensure to keep the primary key name "_id", with the type "ObjectId".'
```
kraenhansen, line 165: In principle I like that we're using the `ObjectId` here, since thats the name of the BSON type - but .. at the same time we're using `object id` a couple of other places ... it feels inconsistent, but perhaps the solution is to lobby that this changes to `ObjectId` in Realm JS?

kraenhansen (changes requested):
Great work! I have a few suggestions related to naming.

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+export class ObjectIdControl extends React.PureComponent<
+  IBaseControlProps,
+  IObjectIdControlState
+> {
+  constructor(props: IBaseControlProps) {
+    super(props);
+
+    this.state = {
+      internalValue: props.value,
```
kraenhansen, line 38: Don't we need `toString` it?

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
kraenhansen: Since this is simply used to grab the initial default value of the property, there is no practical issue here - but we could run into a weird error if we used this component in a context where the property would be updated and we wanted to update our `internalValue` accordingly.

src/ui/RealmBrowser/Content/CreateObjectDialog/index.tsx
steffenagger: True, update incoming. (separating primary key handling from the default if chain, so we fall back if no special-case returns).

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
steffenagger: True. Both namings are leftovers from when the component was a functional component, where value would conflict with the destructuring of props, and the naming was explanatory as to why another handler was needed.

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
steffenagger: I can almost get onboard `handleClearValue`, but I don't think `onClick={this.onClearValue}` makes sense.

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
steffenagger: I don't think so. The external state holds an `ObjectId` where the internal state holds a string that might or might not be a valid 24 char hex string, e.g. `"asd"`.
When the internal state isn't a valid `ObjectId`, we need to reset the external state to have the input "match" the actual state, or at least not match a previous ok state.

I think we should look at getting a more complete form handling library in, reactstrap seems a bit simple.

src/ui/RealmBrowser/AddClassModal/AddClassModal.tsx
steffenagger: It's coming, but might end up being `objectId `, we need to streamline either way.
I'll stick with 'object id', that's pretty ease to find.

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
steffenagger: Argh, I missed the `any` type. I've extended `IBaseControlProps` with an optional generic `ValueType`.. let me know what you think of that solution.

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
kraenhansen: I agree - the `on` prefix is not the best for these type of methods.
I do think however that it would be nice to have consistency in regards to the naming elsewhere, and I've created https://github.com/realm/realm-studio/issues/1296 to help remember that we need to apply that change throughout the codebase.

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
kraenhansen: Okay. It's better to update too much than to have the previous valid value hanging around.
Perhaps it's time for a refactoring of this part of the codebase 👍 (in a separate future PR) - I've created #1297 to track this.

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
```diff
@@ -0,0 +1,102 @@
...
+interface IObjectIdControlState {
+  internalValue: string | null;
+}
+
+export class ObjectIdControl extends React.PureComponent<
+  IBaseControlProps<ObjectId | null>,
+  IObjectIdControlState
+> {
+  state: IObjectIdControlState = {
+    internalValue: this.props.value?.toHexString() ?? null,
```
kraenhansen, line 35: Funky ... I think it's the first time this codebase sees both the optional chaining and the nullish coalescing operator.
... but, there's a first time for everything :) It's less repeated and removes the edge-case of internalValue becomming null in the case of .toHexString() returning the empty string (if that can ever happen?).

kraenhansen (approved):
Looking good!

src/ui/RealmBrowser/Content/CreateObjectDialog/types/ObjectIdControl.tsx
steffenagger: The nullish coalescing operator is just to convert the potential `undefined` value returned from optional chaining, back into `null` (if `value` is `null`).
If `toHexString` returns an empty string (don't know if it can either...), then the empty string would be used, `??` only checks for `undefined` & `null`, otherwise we should use the "good" old `||`. But the control should handle both `ObjectId` & `null` in the case of an optional property.

PR #1305, opened 2020-06-03
Update CHANGELOG.md

bmunkholm:


CHANGELOG.md
```diff
@@ -1,410 +1,31 @@
...
 
 All notable changes to this project will be documented in this file.
-
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
-## Release 10.0.0 (2020-06-03)
-
-[Changes since v10.0.0-mongodb.0](https://github.com/realm/realm-studio/compare/v10.0.0-mongodb.0...v10.0.0)
+## Release 10.0.0-beta.1 (2020-06-03)
```
kraenhansen, line 7: Should this be beta.1? I'm currently suggesting that we release this as v10.0.0, without a prerelease-id but on a separate channel to avoid v3 upgrading to it.
If you're sure about this, we need to update package.json and package-lock.json accordingly.

kraenhansen (changes requested):

CHANGELOG.md
```diff
@@ -1,410 +1,31 @@
...
 
-### Breaking Changes
+This is the first release for MongoDB Realm to be used with the v10.0.0-beta versions of the Realm SDKs.
+NOTE: This version do not support Realm Cloud and Realm Object Server. Realm Studio 3.11 is the latest
+release that supports that.
 
+### Breaking Changes (compared to Studio 3.11)
```
kraenhansen, line 13: Should we mention breaking the ability to use a Realm file which has been opened in an app which uses an older SDK?

bmunkholm: Not sure I understand what you mean. We can open files from an older SDK.

kraenhansen: After releasing v3.11.0 we discussed that this should have probably had been a major version increment, because we considered it a breaking change that older SDKs could no longer open files that had been opened with this new version of Studio.

bmunkholm: ah, ok got it.

PR #1339, opened 2020-10-15
Notarize app (mac)

steffenagger:
### CHANGES:
This PR adds `electron-notarize` & uses it to notarize the mac .app.
This requires a `hardenedRuntime` - configs updated accordingly (as far as I can determine).
A bug in `electron-builder` makes the zipped output useless with no easy workaround, so zipped releases for mac has been disabled (for now).

### TODOS:
- [x] Update changelog.
- [x] Update [realm@10.0.0 PR](https://github.com/realm/realm-studio/pull/1340) & rebase on it.


package.json
```diff
@@ -28,12 +28,13 @@
       "!node_modules/realm/src${/*}",
       "!node_modules/realm/vendor${/*}"
     ],
+    "afterSign": "./scripts/afterSign",
     "mac": {
-      "hardenedRuntime": false,
+      "hardenedRuntime": true,
+      "gatekeeperAssess": false,
```
kraenhansen, line 34: Any reason this is `false`? Also - it looks to me as if this is `false` by default? https://www.electron.build/configuration/mac

scripts/afterSign.js
```diff
@@ -0,0 +1,11 @@
...
+
+exports.default = async (context) => {
+  console.log("electron-builder.afterSign hook triggered.");
+
+  switch (context.electronPlatformName) {
+    case "darwin":
+      await notarizeMacApp(context);
+      break
+  }
+};
```
kraenhansen, line 11:
```suggestion
};

```

scripts/mac/notarize.js
```diff
@@ -0,0 +1,49 @@
...
+
+  // Sanity check
+  if (!fs.existsSync(appPath)) {
+    throw new Error(`File not found: ${appPath}`);
+  }
+
+  await notarize(options);
+
+  console.log(`Done notarizing "${appBundleId}".`);
+};
```
kraenhansen, line 49:
```suggestion
};

```

kraenhansen (approved):

package.json
steffenagger: Yes it's now `false` by default, but if true, `electron-builder` will test prematurely. (it could be omitted if preferred?)

PR #1382, opened 2021-01-27
Reading schema from Realm file when importing CSV (v10)

### LINKED ISSUE #1372: Must be of type 'string?', got 'number' - import failed due to database features
**Versions:**
 - OS: MacOS 10.12 / Linux Mint 20
 - Studio Version - 3.10.1

Hello

I need to import a database into a single MacOS app.

realm opens in Realm Studio-3.10.1, in MongoDB requires an upgrade.

first I need to do an export to view the csv structure, which I was only able to do with Realm Browser 3.0.1

Next, I decided to check whether the macOS app will accept the database I changed.
to do this, I decided to import the csv I just received after exporting, to the same database, deleting the values so that there was no conflict.
Realm Browser 3.0.1 does not import to the same database, but only allows you to create a new database from csv, but The MacOS app does not accept a new database.

then there is only Realm Studio-3.10.1. but it has another problem - it just does not import data, the problem is in the screenshot.

Well, what's the difference that there are numbers for "string", this is not an error, this is a feature of this database.

is it possible to ignore such warnings in Realm Studio-3.10.1? because this is not an error, but Realm Studio-3.10.1 does not allow you to import because of it. MongoDB issues the same warning.

Thanks

kraenhansen:
Fixed CSV importing into an existing Realm. The schema declared by the Realm file was not used, but instead generated from the in the CSV file data. This fixes #1372.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -6,7 +6,7 @@
 
 ### Fixed
 
-- None
+- Fixed CSV importing into an existing Realm. The schema declared by the Realm file was not used, but instead generated from the in the CSV file data. ([#1382](https://github.com/realm/realm-studio/pull/1382), since 1.12.0)
 
 ### Internals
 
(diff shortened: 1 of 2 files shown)


derrikF:
in general, everything was broken...

https://github.com/realm/realm-studio/issues/1372#issuecomment-769632139

PR #1388, opened 2021-01-28
Migrated from tslint to eslint

kraenhansen:


steffenagger (approved):
LGTM.
I thinks we should also tackle the `@typescript-eslint/no-explicit-any` warnings, but we should probably align the approach, and do that at a later time.

PR #1392, opened 2021-01-31
CSV import improvements

### LINKED ISSUE #1389: CSV Import - Imports to class, not subclass
**Describe the bug**

On the newest version of Realm Studio, when I click "Import from CSV", the items appear in the superclass instead of the subclass. 

**To Reproduce**
1. Create a class and then subclass in a realm database. 
(For me, the superclass is "Card" and I have various types of cards.)

2. From Realm Studio, go to Import Data From... CSV. Select a CSV file with a name that matches the subclass. The data will be imported into the superclass and not the subclass. 
(In prior versions, I'd select the CSV with the file name that corresponds to my subclass and select import. All data would then be imported into the subclass. Now, regardless of the name of the CSV, the data goes to the superclass and data from the properties specific to my subclasses is lost. 

**Expected behavior**
I expected the data to be imported into the subclass, as was the case in prior versions. 

**Screenshots**

See attached. 
-"Old Version" shows the correct import into Notes (subclass). 
-"New Version" shows the incorrect import into Card (superclass). 

**Versions:**
 - OS: 11.0.1
 - Studio Version:  10.1.1
 - Realm Object Server version [e.g. 3.0.1]

**Additional context**
On new version
On old version

kraenhansen:
This fixes #1389 by adding a prompt for users to choose the destination class name for each file imported into an existing Realm.

Screenshot 2021-02-08 at 15 01 14

### DIFF
--- a/src/services/data-importer/csv/CSVDataImporter.ts
+++ b/src/services/data-importer/csv/CSVDataImporter.ts
@@ -27,8 +27,7 @@ export class CSVDataImporter extends DataImporter {
   private static readonly NUMBER_OF_INSERTS_BEFORE_COMMIT = 10000;
 
   public import(realm: Realm, files: ImportableFile[]) {
-    console.log("import called", realm, files);
-    for(const file of files) {
+    for (const file of files) {
       const className = file.className;
       const schema = realm.schema.find(s => s.name === className);
       if (!schema) {
@@ -95,7 +94,7 @@ export class CSVDataImporter extends DataImporter {
       }
 
       realm.commitTransaction();
-    };
+    }
   }
 
   private convertToType(value: string, type: PropertyType) {
--- a/src/services/data-importer/csv/CSVDataImporter.ts
+++ b/src/services/data-importer/csv/CSVDataImporter.ts
@@ -30,7 +30,9 @@ export class CSVDataImporter extends DataImporter {
       const className = file.className;
       const schema = realm.schema.find(s => s.name === className);
       if (!schema) {
-        throw new Error(`Unable to import ${file.path}: Class name (${className}) missing from schema`);
+        throw new Error(
+          `Unable to import ${file.path}: Class name (${className}) missing from schema`,
+        );
       }
 
       const rawCSV = fs.readFileSync(file.path, 'utf8');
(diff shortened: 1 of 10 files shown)


src/services/data-importer/csv/CSVDataImporter.ts
```diff
@@ -18,60 +18,54 @@
 
 import fs from 'fs-extra';
 import papaparse from 'papaparse';
+import { ObjectSchemaProperty, PropertyType } from 'realm';
 
-import { DataImporter } from '../DataImporter';
+import { DataImporter, ImportableFile } from '../DataImporter';
 
 export class CSVDataImporter extends DataImporter {
   private static readonly NUMBER_OF_INSERTS_BEFORE_COMMIT = 10000;
```
steffenagger, line 26: (I know this isn't a part of the changes, but...) Does batch commits make sense in transactions? I would argue it contradicts transactions.

src/services/data-importer/csv/CSVDataImporter.ts
```diff
@@ -18,60 +18,54 @@
...
-                    throw new Error(
-                      `Can not parse ${row[prop]} as int at line ${lineIndex}`,
-                    );
-                  }
-                  object[prop] = floatNumber;
-                  break;
-                default:
-                  // string?
-                  object[prop] = row[prop];
+              if (propertySchema.optional && dataValue === '') {
```
steffenagger, line 56: Comment: This seems wrong, or at least hard to actually insert empty strings on optional fields, but it seems this is a limitation in the CSV standard.

steffenagger (commented):
Looks good, except `showMessageBoxSync` doesn't really fit the task. Do we have any other usable (custom) popups implemented, we could use?

src/services/data-importer/csv/CSVDataImporter.ts
kraenhansen: Yeah, I agree. There is the edge case of empty strings. Ideally this should be a setting when importing.
Since its possible to go from `null` to an empty string via the UI (although thats 1 operation per row) but not the other way around, I think this is a good default for now (until we decide to invest more in this).

src/services/data-importer/csv/CSVDataImporter.ts
kraenhansen: It makes sense? The implementation opens a write transaction, inserts up to 10000 objects and then commits and opens a new transaction. It's like wrapping every 10000 `create` calls in a `write` call.

steffenagger: Yes, but if you are importing a 20.000 rows csv, would you be interested in just the first 10.000 in the db, if something fails for the remaining 10.000? (Isn't that what potentially could happen?)

kraenhansen: You're right 👍  I think there's a trade-off here, but I can't remember why we shouldn't just make one huge transaction.

steffenagger (approved):
Looks good.
A couple of console.logs left in, and perhaps disable the `import` button, when target classes hasn't been selected?

steffenagger (approved):

PR #1429, opened 2021-05-21
m11: v21 file-format link for demo-file

steffenagger:
"Download a demo Realm file" link pointed to demo-v21.realm.


kraenhansen (approved):
9/10 points.
With a note for the changelog, this would be a 10/10 😎

PR #1452, opened 2021-09-16
Shallow mode of displaying objects and lists of objects

kraenhansen:
This will skip other objects and lists when building the string representation of objects and lists in the browser.

Screenshot 2021-09-16 at 15 00 40

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2017",
+    "target": "es2019",
     "moduleResolution": "node",
     "module": "esnext",
     "esModuleInterop": true,
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -6,7 +6,7 @@
 
 ### Fixed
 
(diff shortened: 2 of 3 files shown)


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2017",
+    "target": "es2019",
```
kraenhansen, line 3: I needed this to use `Object.fromEntries`

kneth (approved):
Will we ever have `shallow` set to true?

kraenhansen:
> Will we ever have `shallow` set to true?

Sure 🙂  https://github.com/realm/realm-studio/pull/1452/files#diff-91414c404fd51d6fb8e47ba511fd1a5c410e5fe3e732254f0f6d621175fcd647R134

I think you meant to say `false`.
But yeah I tried to focus on getting this working and out. I think this part of the code need a bit more love at some point.

takameyer (approved):
LGTM

PR #1572, opened 2023-04-20
Small change to how to detect encrypted Realm files

### LINKED ISSUE #1571: Unable to open encrypted realm in Macos
**Describe the bug**
Unable to open encrypted realm in Macos. Usually it will appear like a modal popup where you can enter the encryption key but right now it straight away show an error.

**To Reproduce**
1. Open Realm Studio and open any encrypted realm you have
2. And later you will see an error

**Screenshots**
If applicable, 
Screenshot 2023-04-20 at 10 52 01 AM

**Versions:**
 - OS: Mac
 - Studio Version: 14.0.0

kneth:
This closes #1571

### DIFF
--- a/src/ui/RealmBrowser/index.tsx
+++ b/src/ui/RealmBrowser/index.tsx
@@ -342,7 +342,7 @@ class RealmBrowserContainer
     const message = err.message || '';
     const mightBeEncrypted =
       message.includes('Not a Realm file.') ||
-      message.includes('Invalid mnemonic') ||
+      message.includes('invalid mnemonic') ||
       message.includes('Realm file initial open failed');
     const realm = this.props.realm;
     if (mightBeEncrypted) {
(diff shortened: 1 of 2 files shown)


kraenhansen (approved):
🙄

src/ui/RealmBrowser/index.tsx
```diff
@@ -342,7 +342,7 @@ class RealmBrowserContainer
     const message = err.message || '';
     const mightBeEncrypted =
       message.includes('Not a Realm file.') ||
-      message.includes('Invalid mnemonic') ||
+      message.includes('invalid mnemonic') ||
```
bmunkholm, line 345: @kneth @kraenhansen What about older files that had it capitalised? That won't work anymore?
So shouldn't you just add the lower case option and not remove the uppercase option?

kneth: The core version as Studio 14 is using, has a different error message than previous versions.

PR #1594, opened 2023-09-07
Fix opening a sync realm that is already open elsewhere

### LINKED ISSUE #1593: Can't open a synchronized Realm open in a different process
The Core error message is `History type not consistent` which is not one of the magic strings we check in https://github.com/realm/realm-studio/blob/b3a348bb866ede5a8eda9ac157c79c725294e328/src/ui/reusable/RealmLoadingComponent/index.tsx#L124.

nirinchev:
Fixes https://github.com/realm/realm-studio/issues/1593

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -6,7 +6,7 @@
 
 ### Fixed
 
-- None
+- Fixed an issue where opening a synchronized Realm that is open in another process would result in an error with a message `History type not consistent`. ([#1593](https://github.com/realm/realm-studio/issues/1593))
 
 ### Internals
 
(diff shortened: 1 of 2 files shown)


RELEASENOTES.md
```diff
@@ -6,7 +6,7 @@
 
 ### Fixed
 
-- None
+- Fixed an issue where opening a synchronized Realm that is open in another process would result in an error with a message `History type not consistent`. ([#1593](https://github.com/realm/realm-studio/issues/1593))
```
elle-j, line 9: Move to CHANGELOG?

elle-j (approved):

RELEASENOTES.md
nirinchev: I think the way Studio works is you add notes to the RELEASENOTES.md which are then moved to CHANGELOG during the release process.

elle-j: Alrighty 🙂 , would have expected it to be the other way around 😛 
(p.s. it had failed `changelog` checks)

nirinchev: Yeah, Studio is the only different one and the check isn't correct for the way the repo is setup.

elle-j: Ah okay I see! 👍

PR #1615, opened 2024-01-03
Add building and signing for all architectures

bwachter:
This adds building, signing and release uploading for all architectures, and prepares adding universal builds for macOS as well as ARM builds for Windows/Linux.

Missing is the slack notification, and the changelog generation - that is mostly copy/paste from other projects, but we probably need to discuss if we should stick that in this workflow as well, or do a separate workflow for that.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,330 @@
+name: Build, sign and publish release
+# this workflow needs the following secrets set up:
+#
+# ARTIFACTORY_PASSWORD  mongoDB artifactory password
+# ARTIFACTORY_USER      matching username
+# GRS_PASSWORD          Password for the Windows/Linux signer
+# GRS_USERNAME          Username for the Windows/Linux signer
+# NOTARY_KEY_ID         Key id for the macOS notary
+# NOTARY_SECRET         Secret for the macOS notary
+# S3_ACCESS_KEY         Access key with write access to the release bucket
+# S3_SECRET_KEY         The matching secret key
+# WG_CONFIG             A complete wireguard configuration file for notary access
+
+on:
+  workflow_dispatch:
+
+env:
+  GRS_CONFIG_USER1_USERNAME: ${{ secrets.grs_username }}
+  GRS_CONFIG_USER1_PASSWORD: ${{ secrets.grs_password }}
+
+jobs:
+  build_sign_lw:
+    name: Build/sign Windows/Linux
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - name: Install wine
+        run: |
+          sudo dpkg --add-architecture i386
+          sudo apt-get update
+          sudo apt install -y wine wine32
+      - name: Install dependencies
+        run: npm ci --legacy-peer-deps
+      - name: Build the bundle
+        run: npm run build
+      - name: Store package version in environment
+        run: |
+          PACKAGE_VERSION=`jq -r '.version' package.json`
+          echo "PACKAGE_VERSION=$PACKAGE_VERSION" >> $GITHUB_ENV
+      # Both the Windows executable as well as the installer need to be signed.
+      # Electron builder will not rebuild existing files, even if they are
+      # different from the build results - that behaviour could be used to have
+      # it stop before packing everything into installers to sign the relevant
+      # binaries in place. Additionally there's also the option to pack up a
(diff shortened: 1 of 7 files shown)


resources/latest-mac.yml
```diff
@@ -0,0 +1,12 @@
...
+  - url: Realm Studio-VERSION-mac.zip
+    sha512: ZIP_SHA
+    size: ZIP_SIZE
+    blockMapSize: ZIP_BLOCKMAP_SIZE
+  - url: Realm Studio-VERSION.dmg
+    sha512: DMG_SHA
+    size: DMG_SIZE
+path: Realm Studio-VERSION.zip
+sha512: ZIP_SHA
+RELEASE_DATE
```
kraenhansen: Was this committed on purpose? If so, what is it used for?

actions/sign-macos/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ubuntu:20.04
+USER root
+
+WORKDIR /macnotary/work
+ARG MACNOTARY_BIN=/macnotary/bin
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN mkdir -p /macnotary/work && mkdir -p /macnotary/bin
+RUN apt-get update && apt-get -y install curl unzip
+RUN curl -LO https://macos-notary-1628249594.s3.amazonaws.com/releases/client/latest/linux_amd64.zip && \
```
kraenhansen, line 10: Is this supposed to be public information or should the URL be placed in a secret?

actions/sign-macos/action.yml
```diff
@@ -0,0 +1,23 @@
...
+    description: output file
+    required: True
+    default: output.zip
+runs:
+  using: 'docker'
+  image: Dockerfile
+  args:
+    - bash
+    - -c
+    - macnotary -f "${{ inputs.binary }}" -m notarizeAndSign -u https://dev.macos-notary.build.10gen.cc/api -s "${{ inputs.secret }}" -k "${{ inputs.notaryKeyId }}" -b com.mongodb.realm-studio -o "${{ inputs.output }}" -t app
```
kraenhansen, line 23: How about this URL, should it be public knowledge?

.github/workflows/build.yml
```diff
@@ -0,0 +1,302 @@
...
+        run: |
+          cd dist/mac && zip --symlinks -r "../../Realm Studio.app.zip" "Realm Studio.app"
+          #cd dist/mac-universal && zip --symlinks -r "../../Realm Studio.app.zip" "Realm Studio.app"
+      - name: Download signer
+        run: |
+          wget https://macos-notary-1628249594.s3.amazonaws.com/releases/client/v3.8.1/darwin_amd64.zip
+          unzip darwin_amd64.zip
+      - name: Run signer
+        run: |
+          ./darwin_amd64/macnotary -f "Realm Studio.app.zip" -m notarizeAndSign -u https://dev.macos-notary.build.10gen.cc/api -s "${{ secrets.NOTARY_SECRET }}" -k "${{ secrets.NOTARY_KEY_ID }}" -b com.mongodb.realm-studio -o "output.zip" -t app -e resources/entitlements.mac.plist
```
kraenhansen, line 195: Ditto my two other comments.

resources/latest-mac.yml
bwachter: This is referenced in the commented block for creating auto-update files for mac in the scenario where electron builder can't be relied on to correctly write that out - not used right now, but I thought it's a good idea to keep that documented if it ever comes up.

.github/workflows/build.yml
bwachter: I'm asking the relevant guys  to be sure - my impression so far was that it doesn't have to be secret. At least the signer URL is already referenced by other mongodb repositories.

resources/latest-mac.yml
kraenhansen: In that case, I'd prefer if this was put into a markdown file with an introduction :+1:

.github/workflows/build.yml
kneth: the alternative is to create a secret

kneth (approved):
LGTM

Thank you for documenting it with many comments. Should we update `docs/RELEASING.md`?

resources/latest-mac.yml
bwachter: Probably best would be sticking it into the comment for that code block, with annotation where it needs to go if that is needed again?

actions/sign-macos/Dockerfile
bwachter: I got confirmation that those URLs can be public.

kraenhansen (approved):

PR #1622, opened 2024-02-06
Add prepare workflows and publish github release

bwachter:


kneth (approved):

.github/workflows/build.yml
```diff
@@ -328,3 +332,51 @@ jobs:
...
+          delete-branch: true
+          base: main
+          commit-message: Prepare for vNext
+      - name: Merge Pull Request
+        uses: juliangruber/merge-pull-request-action@333730196b34b74936aad75a4e31c23a57582d14
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          number: ${{ steps.vnext-pr.outputs.pull-request-number }}
+          method: squash
+      - name: 'Post to #realm-releases'
```
kraenhansen: This is not the channel that it'll post in anyway.

```suggestion
      - name: 'Post to Slack'
```

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Update Changelog
+        id: update-changelog
+        uses: realm/ci-actions/update-changelog@fa20eb972b9f018654fdb4e2c7afb52b0532f907
+        with:
+          changelog: ${{ github.workspace }}/CHANGELOG.md
+          version: ${{ inputs.version }}
+      - name: Update package.json and package-lock.json
+        run: |
+          sed -i 's/"version": .*/"version": "${{ steps.update-changelog.outputs.new-version }}",/' package.json
+          npm ci --legacy-peer-deps
```
kraenhansen, line 28: Why the legacy-peer-deps flag?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,44 @@
...
+          npm i --package-lock-only
```
kraenhansen, line 29: Why is this needed after the "npm ci"?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      # The current realm studio release process keeps changes for the upcoming release
```
kraenhansen: I think this should go into the PR description instead of a comment this workflow.

.github/workflows/build.yml
```diff
@@ -328,3 +332,51 @@ jobs:
...
+          name: ${{ steps.get-version.outputs.version }}
+          commit: main
+          tag: ${{ steps.get-version.outputs.version }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          draft: false
+      - name: Update Changelog
+        run: |
+          echo "## vNext (TBD)
+
+          ### Enhancements
```
kraenhansen, line 349: I'm surprised this won't add a lot of indention into the changelog 🤔

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+## vNext (TBD)
+
+### Enhancements
+* None
+
+### Fixed
+* None
+
+### Internals
+* Foo bar baz
```
kraenhansen: Was this ment to be committed?

.github/workflows/prepare-release.yml
kraenhansen: In general, it would be great with a bit of motivation and tradeoffs considered to be added to the description of this PR, if not for us then for any future maintainers.

.github/workflows/prepare-release.yml
bwachter: I was following arguments in the Jenkinsfile

.github/workflows/prepare-release.yml
bwachter: I had problems with package-lock.json not getting updated in my runs - not entirely sure why.

CHANGELOG.md
bwachter: No, leftover from testing if it correctly picks up changes

.github/workflows/build.yml
bwachter: Indent up until the beginning of the echo is syntactical yaml indent, which is stripped off before this gets thrown into the shell - so for the shell that block doesn't have any indents at all.

.github/workflows/prepare-release.yml
bwachter: correct, initially I was going for offering both old and new changelog style in the PR, with comments explaining the difference - but on the way figured it's not worth the effort do add the legacy stuff, and just kept that bit - at which point the comment no longer made sense.

.github/workflows/prepare-release.yml
kraenhansen: I'd appreciate we found the root cause of that.

.github/workflows/build.yml
kraenhansen: Awesome, I didn't know that.

.github/workflows/prepare-release.yml
kneth: Is it related to https://github.com/npm/cli/issues/5113?

PR #1628, opened 2024-02-27
Removing touch of binding.gyp from build.yml

kraenhansen:
We can avoid this, since realm@12 is now released with this file already added: https://github.com/realm/realm-js/blob/main/packages/realm/binding.gyp


bwachter:
Makes sense, this was brought up as one of the possible solutions for the crashes on mac - but I'm not sure if at that time it really still was needed. Either way, it'll be easily discoverable from history in case it shows up again.

PR #1629, opened 2024-02-28
Update build workflow and remove unused scripts

kraenhansen:
This updates the build workflow with a couple of fixes to issues discovered while publishing and removes scripts that are no longer used as we've transitioned to GitHub actions.


.github/workflows/build.yml
```diff
@@ -28,14 +34,14 @@ jobs:
...
         with:
-          node-version: 20
+          node-version: 18
       - name: Install wine
         run: |
           sudo dpkg --add-architecture i386
           sudo apt-get update
           sudo apt install -y wine wine32
       - name: Install dependencies
-        run: npm ci --legacy-peer-deps
```
kraenhansen, line 38: Seems this isn't necessary after all.

.github/workflows/build.yml
```diff
@@ -52,7 +58,12 @@ jobs:
       # Building for additional architectures works without conflicts - so to
       # enable arm64 builds for Windows and Linux just add --arm64 here
       - name: Prepare installer contents
-        run: npx electron-builder -lw --publish never --dir --x64
+        run: |
+          npx electron-builder -lw --publish never --dir --x64
+          # Copying over the app-update.yml from Linux as it is skipped when using --dir
+          # https://github.com/electron-userland/electron-builder/blob/79df54238621fbe48ba20444129950ba2dc49983/packages/app-builder-lib/src/publish/PublishManager.ts#L113-L115
+          cat dist/linux-unpacked/resources/app-update.yml
```
kneth, line 65: Is it needed?

.github/workflows/build.yml
```diff
@@ -28,14 +34,14 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
-          node-version: 20
+          node-version: 18
```
kraenhansen, line 37: Aligning with the version used by the Electron dependency.

kneth (approved):
Bye, bye, Jenkins

.github/workflows/build.yml
kraenhansen: Not strictly needed, but I wanted to be sure it exists (as it's relying on undocumented behaviour of electron-builder) and displaying its contents might help debugging future issues.

PR #1645, opened 2024-04-19
RSTUDIO-534: Restore ability to open a synchronized Realm

### LINKED ISSUE #1644: Realm Studio 15 can't open a synchronized Realm
Due to a breaking change in Realm JS 12, we can no longer pass `true` to the `sync` field of the config when opening a Realm. Instead, we need to pass `openSyncedRealmLocally: true`.

nirinchev:
Fixes https://github.com/realm/realm-studio/issues/1644


src/ui/reusable/RealmLoadingComponent/index.tsx
```diff
@@ -119,10 +119,10 @@ export abstract class RealmLoadingComponent<
             path: realm.path,
             encryptionKey: realm.encryptionKey,
             disableFormatUpgrade: realm.enableFormatUpgrade ? false : true,
-            sync: realm.sync as any,
+            openSyncedRealmLocally: realm.sync,
             schema,
             schemaVersion,
-          });
+          } as any);
```
kraenhansen, line 127: Could we keep this a bit tighter?

```suggestion
          } satisfies Realm.Configuration & Record<string, unknown>);
```

kraenhansen (approved):

PR #1651, opened 2024-06-06
RSTUDIO-528: Update version to add support for collections in mixed

gagik:


package.json
```diff
@@ -197,12 +197,12 @@
     "mocha": "^10.4.0",
     "mocha-github-actions-reporter": "^0.3.1",
     "mocha-junit-reporter": "^2.2.1",
-    "node-sass": "^9.0.0",
     "null-loader": "^4.0.1",
     "prettier": "^3.2.5",
     "react-hot-loader": "^4.13.0",
     "remark": "^15.0.1",
     "resolve-url-loader": "^5.0.0",
+    "sass": "^1.77.4",
```
gagik: `node-sass` is [deprecated](https://github.com/sass/node-sass/?tab=readme-ov-file#node-sass) and was breaking CI lint because of this issue for one of its dependencies: https://github.com/nodejs/node-gyp/issues/2869
`sass` is the newer, active project and seems like a drop-in replacement

.github/workflows/prs-lint-build-package.yml
```diff
@@ -15,6 +15,10 @@ jobs:
       uses: actions/setup-node@v1
       with:
        node-version: 18
+    - name: Setup Python 3.11
+      uses: actions/setup-python@v4
+      with:
+          python-version: '3.11.x'
```
gagik, line 21: CI seems to have updated to python `3.12` by default. node-sass is [deprecated](https://github.com/sass/node-sass/?tab=readme-ov-file#node-sass) and was breaking CI lint because one of its dependencies was using a python standard library which was removed in 3.12: https://github.com/nodejs/node-gyp/issues/2869

So we can...
a) stick to Python 3.11, which I am going to do in this PR for this to be as little breaking as possible.
b) get `sass` (`node-sass` replacement) working in a different issue/PR, which requires some small changes in our current SASS code which has deprecrated fields.

kneth: Please create an issue for option b)

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## vNext (TBD)
+
+### Enhancements
+* Added support for collections in mixed.
```
kneth:
```suggestion
* Added support for displaying collections in mixed.
```

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+
+### Fixed
+* None
+
+### Internals
+* Upgraded Realm to `v12.9.0`. ([#1651](https://github.com/realm/realm-studio/pull/1651)).
```
kneth:
```suggestion
* Upgraded Realm to `v12.9.0`. ([#1651](https://github.com/realm/realm-studio/pull/1651))
```

kneth (approved):

.github/workflows/prs-lint-build-package.yml
kraenhansen: I suspect this too is related to `macos-latest` now being arm64 and missing prebuilds for `node-sass` starting a build from source.

gagik: @kraenhansen I think it's related in sense that the `arm64` might be shipping with a more updated Python version by default but the error I was getting was specifically related to distutils not being in Python 3.12, as seen here: https://github.com/nodejs/node-gyp/issues/2869 rather than binaries

kraenhansen: When installing `node-sass` on the x86 MacOS runner, it doesn't need Python at all, because it downloads a prebuilt binary instead of building from source (requiring Python when running `node-gyp`).

gagik: ah, interesting, I assumed it always depended on Python, good to know! Another reason for the new `sass` as this should be much more compatible from what I have seen.

PR #1654, opened 2024-06-06
RSTUDIO-538: Switch to the sass library

### LINKED ISSUE #1652: Use the new sass library instead of node-sass
**Is your feature request related to a problem? Please describe.**
CI seems to have updated to python `3.12` by default. `node-sass` is [deprecated](https://github.com/sass/node-sass/?tab=readme-ov-file#node-sass) and was breaking CI lint because one of its dependencies was using a python standard library which was removed in 3.12: https://github.com/nodejs/node-gyp/issues/2869. Right now the issue is "fixed" by setting our Python version to `3.11` but it'd be good to have latest `sass` installed which removes the need for this.

`sass` will also help us ensure better compatibility with the `arm64` runner.

**Describe the solution you'd like**

1. Remove `node-sass`.
2. Remove the `Setup Python 3.11` from the linting GitHub Actions workflow.
3. Add `sass` as a dependency. 
4. Fix `sass` warnings we have in our repository (mainly related to deprecated SASS properties).

**Describe alternatives you've considered**
We can also just continue using `node-sass` and `Python 3.11` or wait for a potential update of the library.

**Additional context**
See https://github.com/realm/realm-studio/pull/1651

gagik:
Changes the deprecated `node-sass` dependency and uses https://github.com/sass/migrator#readme to mitigate warnings from the migration, namely: https://sass-lang.com/documentation/breaking-changes/slash-div/

Resolves https://github.com/realm/realm-studio/issues/1652


src/ui/ConnectToServer/ConnectToServer.scss
```diff
@@ -70,7 +70,7 @@
     top: -$spacer;
 
     &__Btn {
-      margin: 0 ($spacer / 4);
+      margin: 0 (calc($spacer / 4));
```
kraenhansen: Using [calc](https://developer.mozilla.org/en-US/docs/Web/CSS/calc) makes this compute when evaluating the CSS rather than when compiling the SASS and adds verbosity. What's the context on these changes?

kneth (approved):

src/ui/ConnectToServer/ConnectToServer.scss
gagik: It's related to this: https://sass-lang.com/documentation/breaking-changes/slash-div/
`math.div` may be a more equivalent change, it's just required imports.
Actually this let me notice that they have an automatic migrator, so I could run that and see what the output is

src/ui/ConnectToServer/ConnectToServer.scss
```diff
@@ -70,7 +70,7 @@
     top: -$spacer;
 
     &__Btn {
-      margin: 0 ($spacer / 4);
+      margin: 0 ($spacer * 0.25);
```
gagik: Interestingly enough, this is what the sass migrator tool does for this: https://github.com/sass/migrator#readme. Seems like the way to go

@kraenhansen It is a lot less verbose now, what do you think?

src/ui/ConnectToServer/ConnectToServer.scss
gagik: `sass` does actually simplify `calc` expressions to a single number in most cases as well although it's a bit different from plain `/`

kraenhansen (approved):

PR #1659, opened 2024-06-07
Build only Mac x64 for the current releases

gagik:
Mac runner because of the change defaults to `arm64`. We may want to support `universal` builds in the future but this will require additional setup.


kraenhansen (commented):
Looks about right to me 👍

PR #1663, opened 2024-06-11
RSTUDIO-539: Better visualization for mixed and dictionaries

### LINKED ISSUE #1662: Better visualization for mixed and dictionaries
**Is your feature request related to a problem? Please describe.**
Since we're using flatted , the stringified output can be weird to read as a human; for example:
`flatted.stringify([123]) == "[[123]]"` or `flatted.stringify([[123]]) == `[["1"],[123]]`

**Describe the solution you'd like**
Use `util.inspect` to visualize these cases instead.

gagik:
Closes https://github.com/realm/realm-studio/issues/1662.

## After:
Screenshot 2024-06-11 at 17 01 31
Screenshot 2024-06-11 at 17 01 51

## Before: 
Screenshot 2024-06-11 at 16 59 31
Screenshot 2024-06-11 at 16 59 38

## Data Used
```jsx
class Car extends Realm.Object {
  static schema = {
    name: "Car",
    properties: {
      car: "Car?",
      mixedField: "mixedField",
      hashMapField: "mixed{}",
    },
  };
}
...
realm.write(() => {
  realm.deleteAll();

  new Car(realm, { mixedField: [[123]], car: null });
  const car = new Car(realm, {
    mixedField: [],
    car: null,
  });
  new Car(realm, {
    mixedField: [[123]],
    car: null,
    hashMapField: { test: "nice", a: { b: "whatever" }, c: [1, 2, 3, {}] },
  });
  new Car(realm, { mixedField: [[123]], car: null });

  new Car(realm, { mixedField: [[123], car], car });
});
```


src/utils/json.ts
```diff
@@ -16,6 +15,24 @@ type SafeJsonOptions = {
...
 };
 
+export const prettifiedInspect = (
+  object: unknown,
+  options?: InspectOptions,
+) => {
+  // If it is possible to serialize the object to a simpler structure with toJSON, do it.
+  const simplifiedObject =
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (object as any).toJSON != null ? (object as any).toJSON() : object;
```
gagik: this is a bit hacky but helps clean things up... mainly symbols appearing in the output.

src/utils/json.ts
```diff
@@ -1,6 +1,5 @@
 import { display as displayDataCell } from '../ui/RealmBrowser/Content/Table/types/DataCell';
-import { stringify } from 'flatted';
-
+import { InspectOptions, inspect } from 'util';
```
kraenhansen, line 2:
```suggestion
import { InspectOptions, inspect } from 'node:util';
```

src/utils/json.ts
kraenhansen: I'd be interested to learn what symbols are appearing, we might want to fix that by ensuring these aren't defined as enumerable by the SDK.

src/utils/json.ts
gagik: Is there an advantage to this import over the other, is it to make it explicit this is from part of node? We actually also import `util` directly in an earlier file (which I chose to leave untouched to avoid unnecessary side effects) so I could also change it there
https://github.com/realm/realm-studio/blob/48b64e14733c8848db191436cdb9e8ea76bf2fda/src/ui/RealmBrowser/display.ts#L19

kneth (approved):
It is more human readable.

src/utils/json.ts
kraenhansen: > Is there an advantage to this import over the other, is it to make it explicit this is from part of node?

Yes. The most obvious is to avoid loading a `node_modules/util` which might have been installed by some mis-behaving dependency.

> We actually also import `util` directly in an earlier file

It's most likely because it predates the introduction of the `node:` prefix for the built-in modules.

src/utils/json.ts
kraenhansen: If you're interested in a less hack'y way for doing this:

```typescript
const simplifiedObject = object && typeof object === "object" && "toJSON" in object && typeof object.toJSON === "function";
```

https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEYD2A7AzgF3kgRgK3AwC54BXFAaxSQHcUBuAKEeXSzQEsBbABwg4BmHEMADy+QvAC82CWCwAyBfAwBPHiCQDZBedKkyARLl0ZD8JfEMYkAKQDKogHLmOKHZMtqNWj-IB0Ng7O+kYC5PIcqIZMQA

kraenhansen (commented):
👍

PR #1666, opened 2024-06-20
RSTUDIO-540: Fix connect to server

### LINKED ISSUE #1665: Can't open device sync in realm studio?
Hello,

I'm having a weird case, recently downloaded realm studio 15.20.0 it installed correctly but when I tried to connect into my appID, it throws me this error always:
image

The thing is, that I tried with many differents appIds all with the same outcome: Realm.App._clearAppCache is not a function
I don't know if this is specific because I'm using an M1 Mac?

Thanks in advance

kraenhansen:
This fixes #1665 by removing the call to an internal method which is no longer present in `realm@12`.


src/ui/ConnectToServer/index.tsx
```diff
@@ -35,7 +35,8 @@ import {
 
 import { ConnectToServer } from './ConnectToServer';
 
-const DEFAULT_BASE_URL = 'https://realm.mongodb.com';
+const DEFAULT_BASE_URL = 'https://services.cloud.mongodb.com';
+const LEGACY_BASE_URL = 'https://realm.mongodb.com';
```
kraenhansen, line 39: I went ahead and updated the base URL in the PR as well.

src/ui/reusable/RealmLoadingComponent/index.tsx
```diff
@@ -169,6 +170,7 @@ export abstract class RealmLoadingComponent<
                 }
               },
             },
+            */
```
kraenhansen, line 173: Disabling the initial subscription, since this could potentially result in a huge download that we don't want to wait for.

src/ui/ConnectToServer/index.tsx
```diff
@@ -114,9 +115,8 @@ class ConnectToServerContainer extends React.Component<
       const { appId, url } = this.state;
       // Use SDK default (passing undefined) on an empty string.
       const baseUrl = url || DEFAULT_BASE_URL;
-      // Clear test state to invalidate the shared App cache as a work around for
+      // TODO: Clear test state to invalidate the shared App cache as a work around for
```
kraenhansen, line 118: From my testing, I'm unable to reproduce this and it seems fair to disable the workaround for now. I'm leaving the comment for a future where this resurfaces.

src/ui/reusable/RealmLoadingComponent/index.tsx
```diff
@@ -156,11 +156,12 @@ export abstract class RealmLoadingComponent<
         });
         const credentials = hydrateCredentials(realm.credentials);
         const user = await app.logIn(credentials);
-        return new Realm({
+        return Realm.open({
```
kraenhansen, line 159: Using `Realm.open` here to ensure the schema gets downloaded before the Realm is considered loaded.

gagik (approved):

